PR #11, opened 2023-10-03
Support chain configuration

Siegrift:
Closes https://github.com/api3dao/airseeker-v2/issues/3

This PR is largely inspired by https://github.com/api3dao/signed-api/tree/main/packages/pusher and some of the code is copied over (will be later moved to commons repo). The chain configuration is a small subset of the [old airseeker configuration](https://github.com/api3dao/airseeker/blob/74bf6fc721ac8ae469cc372ec17eebbf890b0517/config/airseeker.example.json#L55).

All properties of the configuration should be documented, but there is no implementation yet. Part of the configuration is `__Temporary__DapiDataRegistry` which is a temporary replacement of `DapiDataRegistry` contract. I think the current form is usable, but it can be freely changed later (since that section will ultimately be removed).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,5 @@ dist
 node_modules
 coverage
 .DS_Store
+airseeker.json
+secrets.env
--- a/README.md
+++ b/README.md
@@ -19,8 +19,8 @@ sensitive information inside secrets.
 
 ### `sponsorWalletMnemonic`
 
-The mnemonic of the wallet used to derive sponsor wallets. Sponsor wallets are derived from dAPI name. It is recommended
-to interpolate this value from secrets. For example:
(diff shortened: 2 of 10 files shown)


vponline (approved):
LGTM 👍 

I guess the `loadConfig` function could be something to have in the `common` repo (it would only need to be changed to accept the config zod schema as an input), along with some of the common zod schemas 🤔

Siegrift:
> I guess the loadConfig function could be something to have in the common repo eventually (it would only need to be changed to accept the config zod schema as an input), along with some of the common zod schemas 🤔

Yeah. I will be working on that starting today. One thing to note is that composing the schema may not be 100% compatible if different zod version is used (which can be the case once the schemas are in commons). This sometimes happened with OIS. The solution was to use the same Zod schema version, but it was a bit annoying... (still better then copypasting it each time though imo).

PR #20, opened 2023-10-04
Add deviationThresholdCoefficient config field

### LINKED ISSUE #14: Add global deviationThresholdCoefficient config
Add a global `deviationThresholdCoefficient` in the config to be used in the deviation threshold check. This is needed for backup Airseekers.

vponline:
Closes #14

I suspected we might also make this value an integer, but wasn't sure if we might have use for fractions for some reason

getting started with an easy one 😄

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,3 +97,11 @@ The Airnode address of the beacon.
 `templateId`
 
 The template ID of the beacon.
+
+### `deviationThresholdCoefficient`
+
+The global coefficient applied to all deviation checks. Used to differentiate alternate deployments. For example:
+
+```jsonc
+"deviationThresholdCoefficient": 1,
+```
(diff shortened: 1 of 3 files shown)


aquarat (approved):
No judgement from me 😝 
Only question from me is, is this a base 1 multiplier? so a secondary Airseeker would be `2` for x2 the thresholds? It'd probably be useful to have the ability to set a float for this (which `.number` allows).

Something else that may not have been considered: the current Airseeker config generators also allow for the heartbeat interval to have a margin added to it for secondary deployments - otherwise different Airseekers every HR interval on low volatility feeds try to execute updates at the same time.

vponline:
> Only question from me is, is this a base 1 multiplier? so a secondary Airseeker would be 2 for x2 the thresholds? It'd probably be useful to have the ability to set a float for this (which .number allows).

Yes this is exactly how I understood the purpose of this field

bbenligiray:
I'd say make this optional with a default value of `1`

PR #21, opened 2023-10-04
[WIP] Initial data fetcher and signed data store implementation

aquarat:
This PR adds a signed data fetcher.
I haven't written tests yet as I want to get an opinion if this approach is reasonable.

Closes https://github.com/api3dao/airseeker-v2/issues/6

### DIFF
--- a/src/signed-api-fetch/data-fetcher.ts
+++ b/src/signed-api-fetch/data-fetcher.ts
@@ -1,38 +1,13 @@
 import { go } from '@api3/promise-utils';
 import axios from 'axios';
-import { z } from 'zod';
 import { groupBy } from 'lodash';
-import { BigNumber } from 'ethers';
-import { Config, EvmAddress, EvmId } from '../config/schema';
+import { Config } from '../config/schema';
 import { logger } from '../logger';
+import { AirnodeAddress, DataStore, signedApiResponseSchema, SignedData } from '../types';
+import { localDataStore } from '../signed-data-store';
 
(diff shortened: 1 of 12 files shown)


src/signed-api-fetch/data-fetcher.ts
```diff
@@ -0,0 +1,304 @@
...
+    const url = urls[Math.ceil(Math.random() * urls.length) - 1];
+    logger.debug(`[${whoAmI}] Calling ${url}/${airnodeAddress}`);
+
+    const result = await go(
+      () =>
+        axiosProd({
+          method: 'get',
+          timeout: HTTP_SIGNED_DATA_API_ATTEMPT_TIMEOUT - HTTP_SIGNED_DATA_API_HEADROOM / 2,
+          //url: `${url}/${airnodeAddress}`,
+          url: `${url}/0xC04575A2773Da9Cd23853A69694e02111b2c4182`, // TODO override here so we can better simulate many apis
```
bdrhn9, line 74: I saw it while browsing the repositories the address here is not the address of the node. `0xc52EeA00154B4fF1EbbF8Ba39FDe37F1AC3B9Fd4` this is the one for Nodary Airnode address.

aquarat: Thanks @bdrhn9, this is just for testing, we ideally need a server that serves many airnodes with fresh data 😅

Do you have other airnode addresses on this server?

bdrhn9: > Do you have other airnode addresses on this server? Also it looks like the timestamps don't change.
Yeah there is one belongs to Nodary `0xc52EeA00154B4fF1EbbF8Ba39FDe37F1AC3B9Fd4` and it is being updated freshly.

PR #26, opened 2023-10-09
Gas price

### LINKED ISSUE #10: Implement recommended gas price
https://docs.google.com/presentation/d/1hCt72FKBthEtCTWRQvjfvjctbJz3MLHIooE7rP20eZA/edit?pli=1#slide=id.g282ed5981d5_0_117

TODO: Update the issue description to what exactly needs to be done (the above is a private presentation).

vponline:
Closes #10


src/gas-price/gas-price.ts
```diff
@@ -0,0 +1,185 @@
...
+    gasPriceStore[chainId]!.lastUpdateTimestamp < Date.now() - scalingWindow * 60 * 1_000
+  ) {
+    return multiplyGasPrice(gasPrice, scalingMultiplier);
+  }
+
+  // Check if the multiplied gas price is within the percentile and return the smaller value
+  // TODO should we check for a minimum length of state gas prices used in the calculation?
+  const sanitizedGasPrice = percentileGasPrice && gasPrice.gt(percentileGasPrice) ? percentileGasPrice : gasPrice;
+
+  // const multipliedGasPrice = multiplyGasPrice(sanitizedGasPrice, recommendedGasPriceMultiplier);
```
aquarat, line 150: Is this related to the comment above?

src/gas-price/gas-price.ts
```diff
@@ -0,0 +1,185 @@
...
+  const fetchInterval = /*config.fetchInterval*/ 30 * 1_000;
+
+  if (!gasPriceCollectorInterval) {
+    gasPriceCollectorInterval = setInterval(runGasPriceCollector, fetchInterval);
+  }
+
+  await Promise.all(
+    Object.entries(config.chains).map(([chainId, chain]) =>
+      go(
+        async () =>
```
aquarat, line 167: AFAIK this is no longer necessary because the new go utils returns the error correctly, might be worth double-checking that.

src/gas-price/gas-price.ts
```diff
@@ -0,0 +1,185 @@
...
+          await airseekerV2ProviderRecommendedGasPrice(
+            chainId,
+            // TODO: what to do with many providers?
```
aquarat, line 170: Implement our our Provider that can do agregation 😆

aquarat (approved):
LGTM :)

vponline:
Please re-review based on the new spec 😄 

The changes include:

- No separate interval for the gas collector, it will be ran within the main loop
- Separate logic to fetch new price and clear old prices
- Separate logic to fetch new price and calculate airseekerV2 price (with scaling and sanitization)
- Checking for pending transactions based on datafeedId:value mapping
- Added function to clear datafeedId:value mapping

src/gas-price/gas-price.ts
```diff
@@ -0,0 +1,236 @@
...
+ * @param providerName
+ * @param sanitizationSamplingWindow
+ */
+export const clearExpiredStoreGasPrices = (
+  chainId: string,
+  providerName: string,
+  sanitizationSamplingWindow: number
+) => {
+  // Remove gasPrices older than the sanitizationSamplingWindow
+  gasPriceStore[chainId]![providerName]!.gasPrices = gasPriceStore[chainId]![providerName]!.gasPrices.filter(
```
Siegrift, line 53: I wonder how efficient this will need to be. For now we can leave it as is, but it may be a good idea to put these records in a data structure that supports faster removals.

src/gas-price/gas-price.ts
```diff
@@ -0,0 +1,236 @@
...
+  );
+};
+
+/**
+ * Saves last transaction details into the store.
+ * @param chainId
+ * @param providerName
+ * @param nonce
+ */
+export const setLastOnChainDatafeedValues = (
```
Siegrift, line 64: Why do we need this? I do not see how data feed values are related to the gas price oracle.

src/gas-price/gas-price.ts
```diff
@@ -0,0 +1,236 @@
...
+  }
+
+  if (!gasPriceStore[chainId]![providerName]) {
+    gasPriceStore[chainId]![providerName] = { gasPrices: [], lastOnChainDataFeedValues: {} };
+  }
+};
+
+/**
+ * Saves a gas price into the store.
+ * @param chainId
```
Siegrift, line 30: nit: I am not a fan of JS Doc comments in typescript, since you can see the names and types of the arguments by default. I like the small text description though.

src/gas-price/gas-price.ts
```diff
@@ -0,0 +1,236 @@
...
+/**
+ *  Calculates the gas price to be used in a transaction based on sanitization and scaling settings.
+ * @param chainId
+ * @param providerName
+ * @param rpcUrl
+ * @param gasSettings
+ * @param nonce
+ * @returns {ethers.BigNumber}
+ */
+export const airseekerV2ProviderRecommendedGasPrice = async (
```
Siegrift, line 167: I'd just call this `getAirseekerRecommendedGasPrice`, but I understand why you called it that way.

Siegrift (commented):
The gas oracle implementation looks good to me, but I think the integration with the main loop will make some functions unused (but we can remove them then).

src/gas-price/gas-price.ts
```diff
@@ -0,0 +1,236 @@
...
+
+/**
+ * Saves a gas price into the store.
+ * @param chainId
+ * @param providerName
+ * @param gasPrice
+ */
+export const setStoreGasPrices = (chainId: string, providerName: string, gasPrice: ethers.BigNumber) => {
+  gasPriceStore[chainId]![providerName]!.gasPrices = [
+    { price: gasPrice, timestamp: Date.now() },
```
Siegrift, line 36: When we say `timestamp` we usually refer to UNIX timestamp (in seconds). I'd rename this one to `timestampMs` just to be explicit or convert to seconds.

src/gas-price/gas-price.ts
```diff
@@ -0,0 +1,236 @@
...
+ * @returns {ethers.BigNumber}
+ */
+export const updateGasPriceStore = async (chainId: string, providerName: string, rpcUrl: string) => {
+  const provider = new ethers.providers.StaticJsonRpcProvider(rpcUrl, {
+    chainId: Number.parseInt(chainId, 10),
+    name: chainId,
+  });
+
+  // Get the provider recommended gas price
+  const goGasPrice = await go(async () => provider.getGasPrice(), { retries: 1, attemptTimeoutMs: 2000 });
```
Siegrift, line 127: Not sure how I feel about retrying the get gas price call. It seems a bit arbitrary to me.

src/gas-price/gas-price.ts
```diff
@@ -0,0 +1,236 @@
...
+): Promise<ethers.BigNumber> => {
+  const { recommendedGasPriceMultiplier, sanitizationPercentile, scalingWindow, scalingMultiplier } = gasSettings;
+  // Get the configured percentile of historical gas prices before adding the new price
+  const percentileGasPrice = getPercentile(
+    sanitizationPercentile,
+    gasPriceStore[chainId]![providerName]!.gasPrices.map((gasPrice) => gasPrice.price)
+  );
+
+  const gasPrice = await updateGasPriceStore(chainId, providerName, rpcUrl);
+  if (!gasPriceStore[chainId]![providerName]!.gasPrices) {
```
Siegrift, line 185: This doesn't make sense to me. This only runs if we don't have any `gasPrices` which happens when we couldn't fetch any gas price from a provider. If that happens the `updateGasPriceStore` throws.

Related to the above, the error handling should be clearer (which function is responsible for handling the error).

src/gas-price/gas-price.ts
```diff
@@ -0,0 +1,236 @@
...
+    return multiplyGasPrice(gasPrice, recommendedGasPriceMultiplier);
+  }
+
+  const lastDataFeedValue =
```
Siegrift, line 189: I thought we decided to use nonces to track updates.

.github/workflows/main.yml
```diff
@@ -42,10 +42,31 @@ jobs:
         run: pnpm run prettier:check && pnpm run eslint:check
       - name: Test
         run: pnpm run test
+  
+  test-e2e:
+    name: Test e2e
+    runs-on: ubuntu-latest
+    # Don't run twice for a push within an internal PR
```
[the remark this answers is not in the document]
vponline: It was included in `operations-database` to reduce the GA runtime since we hit the limit a few times. That's probably still the case?

src/gas-price/gas-price.ts
```diff
@@ -0,0 +1,236 @@
...
+  }
+
+  // Check if the multiplied gas price is within the percentile and return the smaller value
+  // TODO should we check for a minimum length of state gas prices used in the calculation?
+  const sanitizedGasPrice = percentileGasPrice && gasPrice.gt(percentileGasPrice) ? percentileGasPrice : gasPrice;
+
+  return multiplyGasPrice(sanitizedGasPrice, recommendedGasPriceMultiplier);
+};
+
+export const runGasPriceCollector = async () => {
```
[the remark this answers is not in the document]
vponline: True! Will remove the loop entirely 😄

test/e2e/gas-price.feature.ts
```diff
@@ -0,0 +1,147 @@
...
+  await wallet.sendTransaction({
+    to: hre.ethers.constants.AddressZero,
+    ...(gasPriceOverride ? { gasPrice: gasPriceOverride } : {}),
+  });
+};
+
+describe('airseekerV2ProviderRecommendedGasPrice', () => {
+  beforeEach(async () => {
+    // Reset the local hardhat network state for each test to prevent issues with other test contracts
+    await hre.network.provider.send('hardhat_reset');
```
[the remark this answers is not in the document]
vponline: It's useful to reset the provider recommended gas price since otherwise it would fluctuate between tests. We also have the [runInBand](https://jestjs.io/docs/cli#--runinband) option set to run them sequentially.

.github/workflows/main.yml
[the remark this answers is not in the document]
Siegrift: Yeah, why not apply it to all tests then?

test/e2e/gas-price.feature.ts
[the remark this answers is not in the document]
Siegrift: Thanks. I forgot we added that in the other PR.

src/gas-price/gas-price.ts
vponline: This is used to set the datafeedId:datafeedValue mapping to check for pending transactions (based on the value and timestamp)

src/gas-price/gas-price.ts
vponline: It's not guaranteed to work if we use an Airseeker with a single sponsorWallet for multiple datafeeds (and Burak requested that it would work for the v1 way in [this thread](https://api3workspace.slack.com/archives/C05TQPT7PNJ/p1697721456428479?thread_ts=1697707242.099579&cid=C05TQPT7PNJ))

src/gas-price/gas-price.ts
vponline: Errors should probably be handled at the `getAirseekerV2ProviderRecommendedGasPrice` level in the main loop, with no `go` around the `getGasPrice` if we don't want to retry there.

src/gas-price/gas-price.ts
```diff
@@ -0,0 +1,205 @@
...
+  timestampMs: number;
+}
+interface GasState {
+  gasPrices: { price: ethers.BigNumber; timestampMs: number }[];
+  lastOnChainDataFeedValues: Record<string, DataFeedValue>;
+}
+
+export const gasPriceStore: Record<string, Record<string, GasState>> = {};
+
+export const initializeGasStore = (chainId: string, providerName: string) => {
```
Siegrift, line 16: I think we should have just a single state, so I'd move merge this logic with `src/state`.

src/gas-price/gas-price.ts
```diff
@@ -0,0 +1,205 @@
...
+  const gasPrice = await updateGasPriceStore(chainId, providerName, rpcUrl);
+
+  const lastDataFeedValue =
+    newDataFeedUpdateOnChainValues &&
+    gasPriceStore[chainId]![providerName]!.lastOnChainDataFeedValues[newDataFeedUpdateOnChainValues.dataFeedId];
+  // Check if the next update is a retry of a pending transaction and if it has been pending longer than scalingWindow
+  if (
+    lastDataFeedValue &&
+    newDataFeedUpdateOnChainValues &&
+    lastDataFeedValue?.value === newDataFeedUpdateOnChainValues.newDataFeedValue.value &&
```
Siegrift, line 187: In this [slack comment](https://api3workspace.slack.com/archives/C05TQPT7PNJ/p1698222551839029?thread_ts=1698220736.112239&cid=C05TQPT7PNJ) I keep wondering about this. 

We will be making new transactions all the time (since the singed data changes asynchronously) and this will not track pending txs correctly. I do not see any other way except relying on wallet nonces or awaiting the transaction promise.

src/gas-price/gas-price.ts
vponline: Done, that was quite a refactor 😄

test/e2e/gas-price.feature.ts
Siegrift, line 1: I really like these test btw.

src/gas-price/gas-price.ts
Siegrift: I guess the important part is that if the on-chain value changes there had to be an update. If that's only what we care about then the logic is sound.

Siegrift (approved):
🙏 Thanks for the discussions and making this work.

With https://github.com/api3dao/airseeker-v2/pull/26/files#r1372062174 I think the logic is now clear to me.

PR #29, opened 2023-10-11
Fetch data from signed APIs

### LINKED ISSUE #6: Fetch data from signed API
There should be a global `fetchInterval` (maybe configurable) and Airseeker should periodically query the signed API for data.

### LINKED ISSUE #19: Implement AirseekerV2's internal state
This task is to define the internal state of the application and methods to get and set it.

### LINKED ISSUE #23: [Backlog/deferred] Signed data checking
Airseeker v1 does not check received signed data and implementing this here is considered to be a low priority. This task is to track eventually implementing signed data signature checks.

aquarat:
This PR adds code that:
- Fetches signed data from a set of signed data API URLs
- Checks the timestamp and signature on each received data sample
- If valid adds the data to a datastore
- Datastore pruning added
- Adds a basic state (cloneDeep a good idea? may impact performance? hopefully not duplicated?)

It was decided in a meeting that the API calls be done all at once/not staggered for simplicity.

I could probably move the current initialisation boilerplate into a `main` file/function.

Note to self: 
- Also check airnode address of received data matches configured airnode address/filter out data that won't be used before storing

Closes #23 
Closes #6 
Closes #19 

Screenshot 2023-10-12 at 11 25 09


src/signed-data-store/signed-data-store.ts
```diff
@@ -0,0 +1,116 @@
...
+        BigNumber.from(encodedValue).div(10e10).toNumber() / 10e8
+      }`
+    );
+    return false;
+  }
+
+  return true;
+};
+
+const setStoreDataPoint = async (signedData: SignedData) => {
```
vponline, line 50: This doesn't need to be `async`, right?
```suggestion
const setStoreDataPoint = (signedData: SignedData) => {
```

src/signed-api-fetch/data-fetcher.ts
```diff
@@ -0,0 +1,200 @@
...
+    )
+  );
+
+  return Promise.allSettled(
+    urls.map((url, idx) =>
+      go(
+        async () => {
+          const payload = await callSignedDataApi(url, idx.toString());
+
+          logErrors(await Promise.allSettled(payload.map(localDataStore.setStoreDataPoint)));
```
vponline, line 100: If `setStoreDataPoint` doesn't need to be async, could use `goSync` instead?

vponline (approved):
Looking greatly simplified 😄, so no more batching, right?

vponline:
I think this is also closes #19

src/signed-data-store/signed-data-store.ts
```diff
@@ -0,0 +1,116 @@
...
+  signedApiStore[airnode]![templateId] = { signature, timestamp, encodedValue };
+};
+
+const getStoreDataPoint = async (airnode: AirnodeAddress, templateId: TemplateId) =>
+  (signedApiStore[airnode] ?? {})[templateId];
+
+const clear = async () => {
+  Object.keys(signedApiStore).forEach((airnodeAddress) => {
+    delete signedApiStore[airnodeAddress];
+  });
```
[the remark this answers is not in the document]
aquarat: It's a `const` 😆  but I'll make it a let 😅

src/state/state.ts
```diff
@@ -0,0 +1,15 @@
...
+import { Config } from '../config/schema';
+
+type State = {
+  config: Config;
+  dataFetcherInterval?: NodeJS.Timeout;
+};
+
+let state: State | undefined;
+
+export const getState = (): State => cloneDeep(state)!;
```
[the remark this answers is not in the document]
aquarat: Yeah, at the start of the app the state is undefined. It could I guess also be initialised with an empty state but that's a lot more verbose than an `!` 🤷  or the state could be initialised as an object and then given as an argument to consuming functions.

src/signed-data-store/signed-data-store.ts
aquarat: No, but I figured, in line with other functions in this repo that we might want to make this async so we can drop in a key/value store. It might be early optimisation but async paths are infectious and changing it later on could be a serious pain.

src/signed-api-fetch/data-fetcher.ts
aquarat: Yes, it could definitely be `goSync`, but see comment above regarding a drop-in async data store.

src/signed-data-store/signed-data-store.ts
aquarat: I should have probably commented why I did it that way :D

src/signed-api-fetch/data-fetcher.ts
```diff
@@ -0,0 +1,200 @@
...
+          retries: 0,
+          totalTimeoutMs: fetchInterval + HTTP_SIGNED_DATA_API_HEADROOM,
+          attemptTimeoutMs: fetchInterval + HTTP_SIGNED_DATA_API_HEADROOM - 100,
+        }
+      )
+    )
+  );
+};
+
+// Everything from this point won't be needed in production.
```
Siegrift, line 112: If you want to leave it in the codebase, I think you should move it to `test` folder.

src/signed-data-store/signed-data-store.ts
```diff
@@ -0,0 +1,114 @@
...
+import { BigNumber, ethers } from 'ethers';
+import { LocalSignedData, SignedData, AirnodeAddress, DataStore, TemplateId } from '../types';
+import { logger } from '../logger';
+
+// A simple in-memory data store implementation - the interface allows for swapping in a remote key/value store
+
+let signedApiStore: Record<AirnodeAddress, Record<TemplateId, LocalSignedData>> = {};
+let pruner: NodeJS.Timeout | undefined;
+
+export const checkMessage = ({ airnode, templateId, timestamp, signature, encodedValue }: SignedData) => {
```
Siegrift, line 11:
```suggestion
export const verifySignedData = ({ airnode, templateId, timestamp, signature, encodedValue }: SignedData) => {
```

src/signed-data-store/signed-data-store.ts
```diff
@@ -0,0 +1,114 @@
...
+  // 'data' is: ethers.utils.defaultAbiCoder.encode(['int256'], [beaconValue]);
```
Siegrift, line 12:
```suggestion
  // 'encodedValue' is: ethers.utils.defaultAbiCoder.encode(['int256'], [beaconValue]);
```

src/signed-data-store/signed-data-store.ts
```diff
@@ -0,0 +1,114 @@
...
+  );
+
+  const signerAddr = ethers.utils.verifyMessage(message, signature);
+  return signerAddr !== airnode;
+};
+
+export const checkSignedDataIntegrity = (signedData: SignedData) => {
+  const { airnode, templateId, timestamp, encodedValue } = signedData;
+
+  if (parseInt(timestamp) * 1_000 > Date.now()) {
```
Siegrift, line 25: I don't think we need this. We trust the pusher (data provider) and if they are malicious they can straight up misreport (but their signature is there so they won't do it). 

I think we can just have the `verifySignedData`

src/signed-data-store/signed-data-store.ts
```diff
@@ -0,0 +1,114 @@
...
+    return;
+  }
+
+  if (!signedApiStore[airnode]) {
+    signedApiStore[airnode] = {};
+  }
+
+  const existingValue = signedApiStore[airnode]![templateId];
+  if (existingValue && existingValue.timestamp >= timestamp) {
+    logger.debug('Dropping sample at data store as not fresh.');
```
Siegrift, line 65: I found the message a bit hard to read. What about?

```suggestion
    logger.debug('Skipping store update. The existing store value is fresher.');
```

src/signed-data-store/signed-data-store.ts
```diff
@@ -0,0 +1,114 @@
...
+
+const init = async () => {
+  pruner = setInterval(prune, 300_000);
+};
+
+const shutdown = async () => {
+  clearInterval(pruner);
+};
+
+export const localDataStore: DataStore = {
```
Siegrift, line 107: Why define it like this instead of putting `exports` before every function? You avoid manually creating the `DataStore` type.

src/state/state.ts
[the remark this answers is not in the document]
Siegrift: I'd avoid clone deep and just do shallow state updates.

> Yeah, at the start of the app the state is undefined.

What I usually do is throw an error in the getter if the value is not defined. The expectation is that the state is initalized at the start of the app, but potentially needs some values (e.g. config).

src/types.ts
```diff
@@ -0,0 +1,34 @@
+import { z } from 'zod';
+import { EvmAddress, evmAddressSchema, EvmId, evmIdSchema } from './config/schema';
+
+export type AirnodeAddress = EvmAddress;
+export type TemplateId = EvmId;
+
+// Taken from https://github.com/api3dao/signed-api/blob/main/packages/api/src/schema.ts
+// TODO should be imported
```
Siegrift, line 8: Honestly, I'd not bother. Making these two services coupled will cause more complexity then advantages.

But I am curious how do we plan to handle compatibility between these systems. That seems out of the scope of this PR though.

src/utils/evm.ts
Siegrift, line 1: This should not belong production code. Maybe `test` folder?

src/signed-data-store/signed-data-store.ts
aquarat: Just noting that we decided in the meeting to remove this 👌

aquarat: Update:
- Mirror what's in the contract
- Log a WARN message if timestamp is in the future.

Siegrift: Yeah, I take it that you will do it in this repo and will on only create an issue in signed-api repo.

src/index.ts
```diff
@@ -1,2 +1,6 @@
+export * from './signed-api-fetch';
```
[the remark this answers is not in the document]
aquarat: I'm happy to remove them 👌 
I assume the Express instance is for health checks?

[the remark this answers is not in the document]
Siegrift: > I assume the Express instance is for health checks?

Actually, no scratch that. I don't think we want any incoming requests in Airseeker (except maybe for healthcheck)

src/signed-data-store/signed-data-store.ts
aquarat: Looks like you've already created the issue, thanks :) https://github.com/api3dao/signed-api/issues/85

src/signed-api-fetch/data-fetcher.test.ts
```diff
@@ -0,0 +1,67 @@
...
+      })
+    );
+
+    const dataFetcherPromise = runDataFetcher();
+
+    await expect(dataFetcherPromise).resolves.toBeDefined();
+
+    await stopDataFetcher();
+
+    expect(setStoreDataPointSpy).toHaveBeenCalledTimes(30);
```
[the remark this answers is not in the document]
aquarat: The original test/mock config generated 30x data samples to fetch, I've reduced it to 2x.

src/signed-api-fetch/data-fetcher.test.ts
```diff
@@ -0,0 +1,67 @@
+import { init, runDataFetcher, setAxios, stopDataFetcher } from './data-fetcher';
+import { localDataStore } from '../signed-data-store';
+
+describe('data fetcher', () => {
+  const mockAxios = jest.fn();
+
+  beforeAll(() => {
+    setAxios(mockAxios);
```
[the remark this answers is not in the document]
aquarat: Unfortunately this approach doesn't work here for some reason, I'm not sure why.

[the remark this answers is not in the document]
Siegrift: Mhh, I'll take a look. There could be something wrong with the setup or something.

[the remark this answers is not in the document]
aquarat: okay I think I've figured it out, but it's quite different to the linked code above :D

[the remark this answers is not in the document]
Siegrift: I see now. The difference is that you call the axios directly. If you call it like this:
```
     // Move URL to first param and call it with the HTTP method
      axios.get(url, {
        timeout: HTTP_SIGNED_DATA_API_ATTEMPT_TIMEOUT - HTTP_SIGNED_DATA_API_HEADROOM / 2,
        headers: {
          Accept: 'application/json',
          // TODO add API key?
        },
```

That way you can treat axios as an object and `spyOn` its properties. I find this approach a bit better, but your works as well so feel free to keep it.

.gitignore
```diff
@@ -12,3 +12,4 @@ coverage
 airseeker.json
 secrets.env
 .eslintcache
+scratch.ts
```
Siegrift, line 15: Can we remove this?

package.json
```diff
@@ -11,7 +11,24 @@
...
   ],
+  "scripts": {
+    "build": "tsc --project tsconfig.build.json",
+    "clean": "rm -rf coverage dist",
+    "dev": "nodemon --ext ts,js,json,env  --exec \"pnpm ts-node src/index.ts\"",
+    "eslint:check": "eslint --report-unused-disable-directives --cache --ext js,ts . --max-warnings 0",
+    "eslint:fix": "pnpm run eslint:check --fix",
+    "prettier:check": "prettier --check \"./**/*.{js,ts,md,json}\"",
+    "prettier:fix": "prettier --write \"./**/*.{js,ts,md,json}\"",
+    "prettier:write": "prettier --write \"./**/*.{js,ts,md,json}\"",
```
Siegrift, line 22: nit: rm since we have the same one called prettier:fix

src/config/schema.ts
```diff
@@ -98,6 +101,7 @@ export const configSchema = z
   .object({
     sponsorWalletMnemonic: z.string().refine((mnemonic) => ethers.utils.isValidMnemonic(mnemonic), 'Invalid mnemonic'),
     chains: chainsSchema,
+    fetchInterval: z.number().positive().optional().default(10),
```
Siegrift, line 104: I think we want to be explicit with all our configuration values, so I'd remove the default value.

src/signed-data-store/signed-data-store.ts
```diff
@@ -0,0 +1,130 @@
+import { clearInterval } from 'node:timers';
+import { BigNumber, ethers } from 'ethers';
+import { logger } from '../logger';
+import type { LocalSignedData, SignedData, AirnodeAddress, TemplateId } from '../types';
+
+// A simple in-memory data store implementation - the interface allows for swapping in a remote key/value store
+
```
Siegrift, line 7: rm \n so that the comment is directly above `signedApiStore`

src/signed-data-store/signed-data-store.ts
```diff
@@ -0,0 +1,130 @@
...
+    ethers.utils.solidityKeccak256(['bytes32', 'uint256', 'bytes'], [templateId, timestamp, encodedValue])
+  );
+
+  const signerAddr = ethers.utils.verifyMessage(message, signature);
+  return signerAddr !== airnode;
+};
+
+const checkTimestamp = ({ timestamp, airnode, encodedValue, templateId }: SignedData) => {
+  if (Number.parseInt(timestamp, 10) * 1000 > Date.now() + 60 * 60 * 1000) {
+    logger.warn(
```
Siegrift, line 24: This doesn't matter now, but in the future prefer structured logging. E.g:
```
logger.warn('Refusing to store sample as timestamp is more then one hour in the future', {airnode, templateId, systemDateNow, signedDataDate})
```

src/signed-data-store/signed-data-store.ts
```diff
@@ -0,0 +1,130 @@
...
+      ).toLocaleDateString()} vs now ${new Date().toLocaleDateString()}), ${
+        BigNumber.from(encodedValue).div(10e10).toNumber() / 10e8
+      }`
+    );
+  }
+
+  return true;
+};
+
+export const checkSignedDataIntegrity = (signedData: SignedData) => {
```
Siegrift, line 47: Some of these functions are called `verifyXyz` and some `checkXyz`. It's better to stick to one. I personally like "verify" more.

src/signed-data-store/signed-data-store.ts
```diff
@@ -0,0 +1,130 @@
...
+  const { airnode, templateId, timestamp, encodedValue } = signedData;
+
+  if (!checkTimestamp(signedData)) {
+    return;
```
Siegrift, line 51: I guess you want to `return false` here as well.

src/signed-data-store/signed-data-store.ts
```diff
@@ -0,0 +1,130 @@
...
+export const getStoreDataPoint = async (airnode: AirnodeAddress, templateId: TemplateId) =>
+  signedApiStore[airnode]?.[templateId];
+
+// eslint-disable-next-line @typescript-eslint/require-await
+export const clear = async () => {
+  signedApiStore = {};
+};
+
+// eslint-disable-next-line @typescript-eslint/require-await
+export const prune = async () => {
```
Siegrift, line 107: I don't see this function as useful. There is going to be a single data point for a particular beacon. Even if we have 1000 API providers (airnode addresses) with each having 1000 templates, we are storing 10^6 amount of data.

Burak has a related point about pruning data feed info (airnode address and template ID for beacons) but I similarly think that is not needed.

src/signed-data-store/signed-data-store.ts
```diff
@@ -0,0 +1,130 @@
...
+    `Storing sample for (Airnode ${airnode}) (Template ID ${templateId}) (Timestamp ${new Date(
+      Number.parseInt(timestamp, 10) * 1000
+    ).toISOString()}), ${BigNumber.from(encodedValue).div(10e10).toNumber() / 10e8}`
+  );
+
+  signedApiStore[airnode]![templateId] = { signature, timestamp, encodedValue };
+};
+
+// eslint-disable-next-line @typescript-eslint/require-await
+export const getStoreDataPoint = async (airnode: AirnodeAddress, templateId: TemplateId) =>
```
Siegrift, line 98: I think that the requirements for Airseeker storage are quite small and we can be quite sure to stick with in-memory storage. 

This might not be true for signed API which needs to persist multiple data points for each dAPI (because it can serve delayed data as well) so we opted for "async" in-memory store. But even that is questionable and maybe premature.

src/signed-data-store/index.ts
Siegrift, line 1: I really like the folder structure.

src/types.ts
```diff
@@ -0,0 +1,33 @@
...
+import { type EvmAddress, evmAddressSchema, type EvmId, evmIdSchema } from './config/schema';
+
+export type AirnodeAddress = EvmAddress;
+export type TemplateId = EvmId;
+
+// Taken from https://github.com/api3dao/signed-api/blob/main/packages/api/src/schema.ts
+export const signedDataSchema = z.object({
+  airnode: evmAddressSchema,
+  templateId: evmIdSchema,
+  // beaconId: evmIdSchema, // it is removed prior to tx to us | https://github.com/api3dao/signed-api/blob/b6e0d0700dd9e7547b37eaa65e98b50120220105/packages/api/src/handlers.ts#L94
```
Siegrift, line 11: There is no "tx" to us.

I'd personally remove this comment.

src/types.ts
```diff
@@ -0,0 +1,33 @@
...
+export type SignedData = z.infer<typeof signedDataSchema>;
+
+export const signedApiResponseSchema = z.object({
+  count: z.number().positive(),
+  data: z.record(signedDataSchema),
+});
+
+export type LocalSignedData = Pick<SignedData, 'encodedValue' | 'signature' | 'timestamp'>;
+
+export interface DataStore {
```
Siegrift, line 26: This is unused.

src/utils/evm.ts
```diff
@@ -0,0 +1,59 @@
+import { BigNumber, ethers } from 'ethers';
+import { verifySignedData } from '../signed-data-store';
+import { logger } from '../logger';
+import type { SignedData } from '../types';
+
+export const benchmark = async () => {
```
Siegrift, line 6: This is not production code. Maybe make it a test? 

E.g. running 1K signatures and expecting it to finish in a reasonable amount of time? Not sure about this though since the timing largely depends on the machine capacity.

src/utils/evm.ts
```diff
@@ -0,0 +1,59 @@
...
+  logger.info('Duration per iteration: ', (end - start) / 1000 / iterations);
+  logger.info('Throughput per second: ', 1000 / ((end - start) / iterations));
+};
+
+export const getTestSigner = () =>
+  ethers.Wallet.fromMnemonic('test test test test test test test test test test test junk');
+
+export const signData = async (
+  signer: ethers.Signer,
+  airnode: string,
```
Siegrift, line 40: This parameter is not needed.

src/utils/evm.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  logger.info('Benchmark duration in seconds:', durationMs / 1000);
+  logger.info('Duration per iteration: ', (end - start) / 1000 / iterations);
+  logger.info('Throughput per second: ', 1000 / ((end - start) / iterations));
+};
+
+export const getTestSigner = () =>
+  ethers.Wallet.fromMnemonic('test test test test test test test test test test test junk');
+
+export const signData = async (
```
Siegrift, line 38: You can use this function inside `verifySignedData`

src/utils/evm.ts
```diff
@@ -0,0 +1,59 @@
...
+) =>
+  signer.signMessage(
+    ethers.utils.arrayify(
+      ethers.utils.solidityKeccak256(['bytes32', 'uint256', 'bytes'], [templateId, timestamp, data])
+    )
+  );
+
+export const generateRandomBytes32 = () => ethers.utils.hexlify(ethers.utils.randomBytes(32));
+
+if (require.main === module) {
```
Siegrift, line 53: This should be removed.

test/fixtures/mock-config.ts
```diff
@@ -0,0 +1,63 @@
...
+        },
+        activeDapiNames: [],
+      },
+    },
+  },
+  fetchInterval: 10,
+  deviationThresholdCoefficient: 1,
+});
+
+// this should probably be moved to test fixtures
```
Siegrift, line 48: It is now in fixtures.

test/fixtures/mock-config.ts
```diff
@@ -0,0 +1,63 @@
...
+export const init = () => {
+  const config = getConfig();
+  setState({
+    config,
+  });
+};
+
+if (require.main === module) {
```
Siegrift, line 56: Why do we need this? This doesn't seem right to me.

src/signed-data-store/signed-data-store.ts
aquarat: No problem, I've amended it.

src/signed-data-store/signed-data-store.ts
aquarat: Good point, I hadn't thought about that.
I agree on the second point I think: I think it's quite useful having that data present.

package.json
Siegrift: Yeah, I agree.

aquarat:
One more time @Siegrift 😆

Siegrift (approved):
👍 LGTM

PR #33, opened 2023-10-20
Add logger from commons

Siegrift:
Closes https://github.com/api3dao/airseeker-v2/issues/8

## Rationale

Adds the logger from @api3/commons.The logger accepts some configuration and I've opted in configuration from ENVs. Alternative would be to put the logger configuration to `config.json`, but it creates a wierd dependency of not being able to use logger until the config is loaded (similarly, if config fails to load you can't log it).

The implementation is very similar to [pusher](https://github.com/api3dao/signed-api/tree/main/packages/pusher#configuration).

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,0 +1,4 @@
+LOGGER_ENABLED=true
+LOG_COLORIZE=true
+LOG_FORMAT=pretty
+LOG_LEVEL=info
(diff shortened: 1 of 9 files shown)


aquarat (approved):
LGTM, nice a neat, I didn't see any specific issues.
I found it odd that the imports in `src/env/env.ts` are separated by new lines, but that's super minor.

Siegrift:
> I found it odd that the imports in src/env/env.ts are separated by new lines, but that's super minor.

That is part of ESLint. It takes care of grouping imports and each group is separated by new line.

PR #37, opened 2023-10-23
Create data feed update loop

Siegrift:
Closes https://github.com/api3dao/airseeker-v2/issues/34


src/update-feeds/temporary-contract-mock.ts
Siegrift, line 1: I moved this to a separate file for two reasons:
1. Can be more easily removed once we use the contract
2. Can more easily spy on `getStaticActiveDapis` because spying on functions from the same modules is awkward.

src/config/schema.ts
```diff
@@ -112,8 +114,8 @@ export const configSchema = z
   .object({
     sponsorWalletMnemonic: z.string().refine((mnemonic) => ethers.utils.isValidMnemonic(mnemonic), 'Invalid mnemonic'),
     chains: chainsSchema,
-    deviationThresholdCoefficient: z.number().optional().default(1),
-    fetchInterval: z.number().positive(),
+    fetchInterval: z.number().positive(), // TODO: Rename to signedDataFetchInterval
+    deviationThresholdCoefficient: z.number(),
```
vponline, line 118: hmm the previous field schema was requested [here](https://github.com/api3dao/airseeker-v2/pull/20#issuecomment-1750856113), perhaps we'd want to set them for the other fields too 🤔

src/update-feeds/update-feeds.ts
```diff
@@ -0,0 +1,91 @@
...
+import { range, size } from 'lodash';
+
+import type { Chain } from '../config/schema';
+import { logger } from '../logger';
+import { getState } from '../state';
+import { isFulfilled, sleep } from '../utils';
+
+import { type ActiveDapisBatch, getStaticActiveDapis } from './temporary-contract-mock';
+
+export const startUpdateFeedsLoops = async () => {
```
vponline, line 11: Minor 😁 
```suggestion
export const startUpdateFeedLoops = async () => {
```

src/update-feeds/update-feeds.ts
```diff
@@ -0,0 +1,91 @@
...
+  const state = getState();
+  const {
+    config: { chains },
+  } = state;
+
+  // Start update loops for each chain in parallel.
+  await Promise.all(
```
vponline, line 18: We might want to make this a `Promise.allSettled` just in case something unexpected throws to not interrupt all loops in case it only affect one chain or provider? 🤔 

If you do that, Aaron implemented a [nice utility function to log errors](https://github.com/api3dao/operations-database/blob/main/src/workers/common-chain-utils.ts#L74)

vponline (approved):
Overall LGTM, only a couple minor notes that are mostly preference anyway 🏅

src/config/schema.ts
Siegrift: I see. I didn't know that. My view is that I prefer explicit over implicit for our configuration files (because every property matters). 

But because this was explicitly requested, I will revert this change.

vponline: I think your point is valid, though, so perhaps they can be set to be explicit, but probably requires confirmation with the rest of the team 😄

src/update-feeds/update-feeds.ts
Siegrift: I purposefully used `Promise.all` because there should not be an error thrown.

PR #38, opened 2023-10-24
Put spec to README

Siegrift:
I forgot to put the spec into README. I think we decided on that on the last weekly call.


aquarat (approved):
LGTM 👌

PR #39, opened 2023-10-24
WIP Initial update conditions checking

### LINKED ISSUE #18: Implement update conditions checks
This task is to implement the update conditions checks based on the API and on-chain values.

To be done after #6 #17 #14

aquarat:
Closes #18


Siegrift:
> This is largely a copy of the update conditions check in Airseeker v1 (why change something that works well?)

I agree.

src/condition-check/condition-check.ts
```diff
@@ -14,21 +14,29 @@ export const calculateUpdateInPercentage = (initialValue: ethers.BigNumber, upda
...
   const nums = [...arr].sort((a, b) => {
     if (a.lt(b)) return -1;
     else if (a.gt(b)) return 1;
     else return 0;
   });
 
-  if (arr.length % 2 !== 0) {
-    return nums[mid];
-  }
+  const midNumber = nums[mid];
```
Siegrift, line 24: Can't you just put some `!` ? I don't think any of these errors can actually happen.

vponline (approved):
LGTM 🚀

Siegrift:
I see that this PR is failing the tests. How come you were able to merge it? 👀
Do we have something wrong with the repo settings?

Siegrift:
> I see that this PR is failing the tests. How come you were able to merge it? 👀
Do we have something wrong with the repo settings?

I realized these https://github.com/api3dao/airseeker-v2/blob/main/.github/workflows/main.yml#L54 is the cause. We have different strategy for running tests in airseeker then we have in Airnode (which is fine imo).

PR #43, opened 2023-10-26
Add check for required chains and providers

### LINKED ISSUE #40: Tighten config validations
There should be at least 1 chain in the config and 1 provider per each chain.

vponline:
Closes #40


Siegrift (approved):
👍 LGTM

PR #44, opened 2023-10-26
Add DapiDataRegistry and use it to fetch active dAPIs

Siegrift:
Closes https://github.com/api3dao/airseeker-v2/issues/35

## Rationale

Because the contract is still in a PR, I've added a branch dependency and setup hardhat to build Typechain types and factories. These are then used in the actual implementation.

You need to run `pnpm i && pnpm run build` to install missing dependencies and build the Typechain artifacts.


.gitignore
```diff
@@ -11,4 +11,11 @@ coverage
 .DS_Store
 airseeker.json
 secrets.env
-.eslintcache
\ No newline at end of file
+.eslintcache
+
+# Hardhat files
```
Siegrift, line 16: These were generated by `hardhat --init`. I am not sure if all of them are needed, but it's better to keep them all.

README.md
```diff
@@ -123,6 +123,10 @@ A record of contract addresses used by Airseeker. If not specified, the addresse
 The address of the Api3ServerV1 contract. If not specified, the address is loaded from the Airnode protocol v1
 repository.
 
+###### DapiDataRegistry
```
Siegrift, line 126: For now making it required, because it is not determined where the contract will be deployed (I guess it will stay in `dapi-management` repo).

package.json
```diff
@@ -12,22 +12,32 @@
...
-    "tsc": "tsc --project .",
-    "docker:build": "docker build -t api3/airseekerv2:latest -f docker/Dockerfile .",
-    "docker:run": "docker run -it --rm api3/airseekerv2:latest",
-    "dev:eth-node": "hardhat node"
+    "test": "jest --config=jest-unit.config.js",
+    "tsc": "tsc --project ."
   },
   "devDependencies": {
     "@api3/ois": "^2.2.0",
+    "@ethersproject/abi": "^5.7.0",
```
Siegrift, line 33: I had to manually install `@ethersproject/abi` and `@ethersproject/providers` because I was getting TS errors in Typechain generated files. This may be an issue with how pnpm internal dependency scoping works, but I haven't investigated deeply because this is dev only dependency and a temporary solution (the proper solution would be to import the Typechain artifacts from the published repo).

src/config/schema.ts
```diff
@@ -87,13 +88,15 @@ export const chainsSchema = z.record(optionalChainSchema).transform((chains, ctx
...
+        ...contracts,
         Api3ServerV1: contracts?.Api3ServerV1 ?? references.Api3ServerV1[chainId],
       });
       if (!parsedContracts.success) {
         ctx.addIssue({
           code: 'custom',
           message: 'Invalid contract addresses',
-          path: ['chains', chainId, 'contracts'],
+          // Show at least the first error.
+          path: [chainId, 'contracts', ...parsedContracts.error.errors[0]!.path],
```
Siegrift, line 99: I realized `chains` should not be part of the message. Also it would be nice if we showed what address was wrong and we already have that information.

README.md
```diff
@@ -113,7 +113,7 @@ A record of chain configurations. The record key is the chain ID. For example:
 }
 ```
 
-##### `contracts` _(optional)_
+##### `contracts`
 
 A record of contract addresses used by Airseeker. If not specified, the addresses are loaded from
```
aquarat, line 118: Given the above this may no longer apply (it's no longer optional)?

bbenligiray:
FYI we're still messing with the interface

aquarat (approved):
I can't fault it, this looks very cool and I learnt a few things like `satisfies` :D 

🚀

README.md
Siegrift: Yeah. Thanks for noticing.

package.json
Siegrift: Actually nevermind. There were no issues with pnpm, but a mismatch of ethers v6 and v5 in the hardhat dependencies.

PR #49, opened 2023-10-27
Update chart

### LINKED ISSUE #41: Update flowchart diagram
We have resolved discussions in [the spec](https://docs.google.com/document/d/1x5QBOGII8IUGjtoNR6PVE_UeqEjRQj2u3Ysa1FQkHf0/edit) and there are these changes (at least):
- No gas loop (done as part of the data feed update loop)
- No signed URL loop (done as part of the data feed update loop)
- No fetching on-chain value (done as part of the data feed update loop)

Double check that the diagram is in-sync with the spec.

vponline:
Closes #41


Siegrift (commented):
I see the tests are not working because of https://github.com/api3dao/airseeker-v2/pull/39#issuecomment-1782529499

I think we can squeeze that in here (but feel free to create another PR if you prefer).

airseeker_v2_pipeline.drawio
Siegrift, line 1: Couple of changes:
1. Remove is `TODO: This is not yet finalized` which can be removed. Also it would be better to clarify that the first batch is fetched with the dAPIs count in a single RPC multicall.
2. (nit) Make the rectangle for `clear stored on chain datafeed value from gas price store` same size as others.

Siegrift (approved):
👍 LGTM

PR #50, opened 2023-10-27
Confirm README has all example config keys in it

### LINKED ISSUE #42: Create a test that ensures configuration is documented in README
Parse the README file and expect all configuration keys of the `airseeker.example.json` to be present.

This is a heuristic that should cover most of the cases of forgetting to write documentation.

aquarat:
Resolves:
> Parse the README file and expect all configuration keys of the `airseeker.example.json` to be present.

Closes #42


test/readme.test.ts
```diff
@@ -0,0 +1,40 @@
...
+
+describe('checks README', () => {
+  it('checks that the README contains all configuration keys in airseeker.example.json', () => {
+    const config = loadConfig(path.join(__dirname, '../config'), 'airseeker.example.json');
+    expect(config).toBeDefined();
+
+    const readmeData = readFileSync(path.join(__dirname, '../README.md')).toString();
+    expect(readmeData).toBeDefined();
+
+    // console.log(config);
```
Siegrift, line 24: remove

src/config/index.ts
```diff
@@ -1,20 +1,19 @@
...
+export const loadConfig = (configPath = join(__dirname, '../../config'), airseekerConfigPath = 'airseeker.json') => {
   const rawSecrets = dotenv.parse(readFileSync(join(configPath, 'secrets.env'), 'utf8'));
 
-  const goLoadConfig = await go(async () => {
-    const rawConfig = JSON.parse(fs.readFileSync(join(configPath, 'airseeker.json'), 'utf8'));
+  const goLoadConfig = goSync(() => {
+    const rawConfig = JSON.parse(fs.readFileSync(join(configPath, airseekerConfigPath), 'utf8'));
     const secrets = parseSecrets(rawSecrets);
-    return configSchema.parseAsync(interpolateSecrets(rawConfig, secrets));
+    return configSchema.safeParse(interpolateSecrets(rawConfig, secrets));
```
Siegrift, line 16: Good catch. The config does not need to be parsed asynchronously.

However, you should use the `.parse` instead of `.safeParse` here (the response type is wrong now).

test/readme.test.ts
```diff
@@ -0,0 +1,40 @@
...
+    const keys = extractKeys(config).filter((item) => !z.coerce.number().safeParse(item).success);
+
+    const missingKeys = keys.filter((key) => readmeData.includes(`\n#### ${key}`));
+
```
Siegrift, line 28: This does exact opposite. We want to filter all keys that are not in README data. The interpolation check like this will not work. I would imagine somtehing like this:

```ts
    const readmeHashtagTitles = readmeData
      .split('\n')
      .map((line) => /^#+ `(?<title>.+)`/.exec(line)?.groups?.title)
      .filter((title) => !!title);
    const readmeBacktickTitles = readmeData
      .split('\n')
      .map((line) => /^`(?<title>[^`]+)`/.exec(line)?.groups?.title)
      .filter((title) => !!title);
    const readmeTitles = [...readmeHashtagTitles, ...readmeBacktickTitles];
```

to capture all names starting with hastags or the ones that start just with "\`" (such as "url").

and get missing ones like this:
```ts
    const missingKeys = configKeys.filter(
      (configKey) => !readmeTitles.some((readmeTitle) => readmeTitle!.includes(configKey))
    );
```

but feel free to come up with a nicer heuristic.

test/readme.test.ts
```diff
@@ -0,0 +1,40 @@
...
+describe('checks README', () => {
+  it('checks that the README contains all configuration keys in airseeker.example.json', () => {
+    const config = loadConfig(path.join(__dirname, '../config'), 'airseeker.example.json');
+    expect(config).toBeDefined();
+
+    const readmeData = readFileSync(path.join(__dirname, '../README.md')).toString();
+    expect(readmeData).toBeDefined();
+
+    // console.log(config);
+    const keys = extractKeys(config).filter((item) => !z.coerce.number().safeParse(item).success);
```
Siegrift, line 25: You will need some manual filtering for object keys (e.g. hardhat) so might as well just create a the list of properties to filter out (I think only "31337" and "hardhat").

src/config/index.ts
aquarat: Weirdly it didn't fail, I've swapped it out 👌

test/readme.test.ts
aquarat: Sorry about that, I didn't test the inverse. I've added in the negation and changed the README a bit. If you prefer I can go with your approach above instead 👌

README.md
```diff
@@ -126,75 +126,75 @@ The address of the Api3ServerV1 contract. If not specified, the address is loade
...
 
 A mapping from data feed ID to a list of beacon data.
 
-###### `dataFeedIdToBeacons<DATA_FEED_ID>`
+##### `dataFeedIdToBeacons<DATA_FEED_ID>[n]`
 
 A single element array for a beacon data. If the data feed is a beacon set, the array contains the data for all the
 beacons in the beacon set (in correct order).
 
-`dataFeedIdToBeacons<DATA_FEED_ID>[n]`
```
Siegrift, line 161: I ran out of nesting options so I just put it in backticks. I think what you did (make it title with #####) is better so I'd keep it.

README.md
```diff
@@ -48,30 +48,30 @@ LOG_FORMAT=json
 LOG_LEVEL=info
 ```
 
-#### `LOGGER_ENABLED`
+### `LOGGER_ENABLED`
```
Siegrift, line 51: These titles were nested because the `### Environment variables` already uses `###`.

src/config/index.ts
```diff
@@ -1,20 +1,23 @@
...
 import dotenv from 'dotenv';
 
 import { configSchema } from './schema';
 import { interpolateSecrets, parseSecrets } from './utils';
 
-export const loadConfig = async () => {
-  const configPath = join(__dirname, '../../config');
-  const rawSecrets = dotenv.parse(readFileSync(join(configPath, 'secrets.env'), 'utf8'));
+export const loadConfig = (
+  configPath = join(__dirname, '../../config'),
```
Siegrift, line 11: Do we need these parameters? I think the implementation works even when the values are hadcoded.

README.md
aquarat: No problem, I've reverted most of the README changes.

aquarat:
I've reverted the README changes, implemented your regex based search and mocked the readFileSync calls so the config loader will work with the example config without require extra arguments :) @Siegrift

Siegrift (approved):
👍 LGTM

PR #55, opened 2023-10-30
Use timestamp to check for pending txs

### LINKED ISSUE #52: Adjust pending transaction scaling to use timestamp check
Adjust the pending transaction check in the gas price calculator to be based on last update timestamp instead of on-chain values.

vponline:
Closes #52


Siegrift (approved):
👍 LGTM

PR #56, opened 2023-10-30
Use multicall to fetch dAPI batches

Siegrift:
Closes https://github.com/api3dao/airseeker-v2/issues/46

## Rationale

I first wanted to only fix the issue of multicalling the first batch together with the dAPIs count. I wanted to verify that I implemented the functionality correctly, so I wanted to write an e2e test for this (to test the on-chain integration). I followed the setup implemented by @acenolaza [here](https://github.com/api3dao/dapi-management/pull/3/files#diff-731d7bf404aa3af378fe6396cbdbd4328ee7f69d5723adec8f2ed5fdb99bb44cR60) but the contract we used in Airseeker was a bit outdated so I needed to update it. This required using the multicall everywhere.

### DIFF
--- a/jest-e2e.config.js
+++ b/jest-e2e.config.js
@@ -15,5 +15,6 @@ module.exports = {
   testEnvironment: 'jest-environment-node',
   testMatch: ['**/?(*.)+(feature).[t]s?(x)'],
   testPathIgnorePatterns: ['<rootDir>/.build', '<rootDir>/dist/', '<rootDir>/build/'],
+  testTimeout: 15_000,
   verbose: true,
 };
--- a/src/env/env.ts
+++ b/src/env/env.ts
@@ -9,7 +9,7 @@ let env: EnvConfig | undefined;
 export const loadEnv = () => {
   if (env) return env;
 
-  dotenv.config({ path: join(__dirname, '../.env') });
+  dotenv.config({ path: join(__dirname, '../../.env') });
 
   const parseResult = envConfigSchema.safeParse(process.env);
   if (!parseResult.success) {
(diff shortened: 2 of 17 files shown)


jest-e2e.config.js
```diff
@@ -15,5 +15,6 @@ module.exports = {
   testEnvironment: 'jest-environment-node',
   testMatch: ['**/?(*.)+(feature).[t]s?(x)'],
   testPathIgnorePatterns: ['<rootDir>/.build', '<rootDir>/dist/', '<rootDir>/build/'],
+  testTimeout: 15_000,
```
Siegrift, line 18: This will set 15 seconds timeout for all e2e tests, which should be enough.

src/env/env.ts
```diff
@@ -9,7 +9,7 @@ let env: EnvConfig | undefined;
 export const loadEnv = () => {
   if (env) return env;
 
-  dotenv.config({ path: join(__dirname, '../.env') });
+  dotenv.config({ path: join(__dirname, '../../.env') });
```
Siegrift, line 12: This previously pointed to a non-existent file.

test/setup/contract.ts
```diff
@@ -0,0 +1,370 @@
+import * as abi from '@api3/airnode-abi';
```
Siegrift, line 1: This file is heavily inspired by https://github.com/api3dao/airseeker/blob/main/test/setup/deployment.ts and https://github.com/api3dao/dapi-management/pull/3/files#diff-731d7bf404aa3af378fe6396cbdbd4328ee7f69d5723adec8f2ed5fdb99bb44cR60

vponline (approved):
LGTM, I didn't see any issues 🚀

PR #59, opened 2023-10-31
Immer state

### LINKED ISSUE #51: Improve local state management
Use [immer](https://immerjs.github.io/immer/) to improve local state management

vponline:
Closes #51


src/gas-price/gas-price.test.ts
```diff
@@ -52,17 +54,11 @@ describe('gas price', () => {
...
-        gasPriceStore: {
-          ...state.gasPriceStore,
-          [chainId]: {
-            ...state.gasPriceStore[chainId],
-            [providerName]: { gasPrices: [], lastOnChainDataFeedValues: {} },
-          },
-        },
-      });
+      setState(
+        produce(getState(), (draft) => {
```
Siegrift, line 58: I think this should be an internal-only detail of `setState`. Maybe call it `updateState` and accept a callback.

src/gas-price/gas-price.test.ts
```diff
@@ -53,16 +53,9 @@ describe('gas price', () => {
...
-        gasPriceStore: {
-          ...state.gasPriceStore,
-          [chainId]: {
-            ...state.gasPriceStore[chainId],
-            [providerName]: { gasPrices: [], sponsorLastUpdateTimestampMs: {} },
-          },
-        },
+      updateState((draft) => {
+        draft.gasPriceStore[chainId] = { [providerName]: { gasPrices: [], sponsorLastUpdateTimestampMs: {} } };
+        return draft;
```
Siegrift, line 58: I don't think we nee to return `draft` because the return value of the produce will return the modified state.

Siegrift (changes requested):
Can you write a few unit tests for the `updateState` function? I am interested to see:
* if we need the `return draft` (I think that return is optional).
* whether it correctly updates immutably

Siegrift (approved):
👍 LGTM

PR #60, opened 2023-10-31
Remove temp dapidataregistry

### LINKED ISSUE #57: Remove __Temporary__DapiDataRegistry
The `__Temporary__DapiDataRegistry` in codebase is no longer needed because we integrated the contract and can now be removed (alongside its documentation).

This should have been done as part of https://github.com/api3dao/airseeker-v2/pull/44 but I forgot to do it.

vponline:
Closes #57


src/state/state.ts
```diff
@@ -13,6 +13,7 @@ export interface State {
   dataFetcherInterval?: NodeJS.Timeout;
   gasPriceStore: Record<string, Record<string, GasState>>;
   signedApiStore: Record<AirnodeAddress, Record<TemplateId, LocalSignedData>>;
+  signedApiUrlStore: Record<string, Record<AirnodeAddress, string>>;
```
vponline, line 16: I've moved these to be available in the state, so they chould be stored here when batches are fetched. Let me know if you prefer something else, though.

Siegrift: Seems good to me

Siegrift (approved):
👍 LGTM

PR #61, opened 2023-10-31
WIP Process dapis

### LINKED ISSUE #47: Determine what dAPIs from a batch need to be updated
Each [dAPI batch is processed separately](https://github.com/api3dao/airseeker-v2/blob/469e82d32ab937a25aa739db9b6a3c5094d17f8d/src/update-feeds/update-feeds.ts#L96). In https://github.com/api3dao/airseeker-v2/issues/18 we implemented the checks (based off Airseeker v1).

What we need is for each individual dAPI to determine whether it needs to be updated or not.

aquarat:
This PR is a bit more extensive than the original issue because of a change in the contract.

It:
- Changes the way signed data is indexed in the signed data store to make it easier to look up values for feed definitions received from the contract
- I assumed that `Datafeed` from the contract was the dataFeedId... it isn't, it's actually an abi encoded bytes array containing the airnode address and templateId - this gives the app what it needs to get signed data and from these values this PR also derives the dataFeedIds
- Introduced a `dynamicState`, but given the removal of the static config this is probably redundant, but needs to be stored somewhere, may affect @vponline 

This will probably do for now.

Closes #47


Siegrift (commented):
Can you please rebase and change the PR title?

aquarat:
Starting fresh on new branch due to merge conflicts.

PR #62, opened 2023-11-02
Add coordinator ID

Siegrift:
Closes https://github.com/api3dao/airseeker-v2/issues/45

## Rationale

We can use the new commons API `logger.runWithContext` where you specify metadata that is logged with every log message during the execution of the function.

To verify it works, you can use the e2e tests with logs turned off to see the metadata is indeed logged with every log message:
```
LOGGER_ENABLED=true LOG_LEVEL=debug pnpm run test update
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,7 +61,7 @@
   },
   "dependencies": {
     "@api3/airnode-protocol-v1": "^2.10.0",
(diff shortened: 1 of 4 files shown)


vponline (approved):
Cool! 🚀

PR #64, opened 2023-11-02
Update clear last stored update timestamp in chart

vponline:
Updates the pending transaction check to refer to last update timestamp


aquarat (approved):
LGTM 👍

PR #65, opened 2023-11-02
Process on-chain-received DApi batches

aquarat:
After a lot of rebasing, this PR:

- Modifies recent work around the state to reflect  the chain-agnostic nature of signed data
    - and the fact that the contract can return multiple signed data URLs per single source data feed
- Modifies the signed datastore to be datafeed ID indexed to better reflect how the data is used
- Checks on-chain values against local values and produces a batch of update decisions


src/state/state.ts
```diff
@@ -2,19 +2,37 @@ import type { BigNumber } from 'ethers';
...
+}
+
 export interface State {
   config: Config;
   dataFetcherInterval?: NodeJS.Timeout;
   gasPriceStore: Record<string, Record<string, GasState>>;
-  signedApiStore: Record<AirnodeAddress, Record<TemplateId, LocalSignedData>>;
-  signedApiUrlStore: Record<string, Record<AirnodeAddress, string>>;
+  signedApiStore: Record<DataFeedId, SignedData>;
+  signedApiUrlStore: { url: string; lastReceived: number }[];
```
vponline, line 34: Minor: could rename to `lastReceivedMs` to be in line with the other millisecond fields we have and separate from the on chain timestamps in seconds.

src/update-feeds/update-feeds.ts
```diff
@@ -124,8 +138,82 @@ export const runUpdateFeed = async (providerName: string, chain: Chain, chainId:
...
-  // TODO: Implement.
+
+  // Start by merging the dynamic state with the state
+  updateDynamicState(batch);
+
+  const allFeeds = getFeedsToUpdate(batch);
+
+  // Flow chart calls for clearing stale data
+  // "for each dAPI where an update is not needed" ... "clear stored on chain datafeed value from gas price store"
+  // TODO The fn called has been removed, what replaces it (if any)? - it may require sponsor wallet derivation
```
vponline, line 214: I handled this [here](https://github.com/api3dao/airseeker-v2/pull/68) 😄

test/fixtures/mock-config.ts
```diff
@@ -38,7 +38,8 @@ export const init = (stateOverride?: Partial<State>) => {
     config,
     gasPriceStore: {},
     signedApiStore: {},
-    signedApiUrlStore: {},
+    signedApiUrlStore: [],
     ...stateOverride,
+    dapis: {},
```
vponline, line 43:
```suggestion
    dapis: {},
    ...stateOverride,
```

vponline (commented):
This is looking great! Only some minor comments

src/update-feeds/update-feeds.ts
aquarat: Awesome, I'll remove this comment 👍

vponline: I think the way batches are processed now is ok, we still need the ones not being updated as they're now 😄

src/types.ts
```diff
@@ -22,3 +25,14 @@ export const signedApiResponseSchema = z.object({
 });
 
 export type LocalSignedData = Pick<SignedData, 'encodedValue' | 'signature' | 'timestamp'>;
+
+export interface DataFeedSingle {
```
Siegrift, line 29: I think it would be better to call this `Beacon` and have the `DecodedDataFeed` specified as:
```ts
export interface DecodedDataFeed {
  dataFeedId: string;
  beacons: Beacon[];
}
```

src/utils.ts
```diff
@@ -14,3 +17,17 @@ export function deriveBeaconId(airnodeAddress: string, templateId: string) {
 export function deriveBeaconSetId(beaconIds: string[]) {
   return goSync(() => ethers.utils.keccak256(ethers.utils.defaultAbiCoder.encode(['bytes32[]'], [beaconIds]))).data;
 }
+
+export const generateDataFeedBytesSingle = (dataFeed: DataFeedSingle) =>
```
Siegrift, line 21: This is `deriveBeaconId` function above.

src/utils.ts
```diff
@@ -14,3 +17,17 @@ export function deriveBeaconId(airnodeAddress: string, templateId: string) {
...
+  ethers.utils.defaultAbiCoder.encode(['address', 'bytes32'], [dataFeed.airnodeAddress, dataFeed.templateId]);
+
+export const generateDataFeedBytesMultiple = (dataFeed: DataFeedSingle[]) =>
```
Siegrift, line 24: This is only used for tests, so it should be moved to `test/utils.ts`.

src/utils.ts
```diff
@@ -14,3 +17,17 @@ export function deriveBeaconId(airnodeAddress: string, templateId: string) {
...
+  ethers.utils.defaultAbiCoder.encode(
+    ['address[]', 'bytes32[]'],
+    [dataFeed.map((item) => item.airnodeAddress), dataFeed.map((item) => item.templateId)]
+  );
+
+export const getDeviationThresholdAsBigNumber = (input: number) =>
```
Siegrift, line 30: Move to `test/utils.ts`

src/utils.ts
```diff
@@ -14,3 +17,17 @@ export function deriveBeaconId(airnodeAddress: string, templateId: string) {
...
+  ethers.BigNumber.from(Math.trunc(input * HUNDRED_PERCENT)).div(ethers.BigNumber.from(100));
+
+export const getUnixTimestamp = (dateString: string) => Math.floor(Date.parse(dateString) / 1000);
```
Siegrift, line 33: Unused.

test/e2e/update-feeds.feature.ts
```diff
@@ -10,5 +10,5 @@ it('reads blockchain data correctly', async () => {
 
   await runUpdateFeed(providerName, chain, chainId);
 
-  expect(logger.debug).toHaveBeenNthCalledWith(2, 'Processing batch of active dAPIs', expect.anything());
+  expect(logger.debug).toHaveBeenCalledWith('Fetching first batch of dAPIs batches');
```
Siegrift, line 13: This is not testing what it should. Because we don't have the full logic implemented, I added a log statement that calls the `processBatch` which means that the blockchain state was read (and decoded) correctly.

src/update-feeds/update-feeds.ts
```diff
@@ -60,9 +67,13 @@ export const runUpdateFeed = async (providerName: string, chain: Chain, chainId:
         await dapiDataRegistry.callStatic.tryMulticall([dapisCountCall, ...readDapiWithIndexCalls])
       );
 
-      const dapisCount = decodeDapisCountResponse(dapiDataRegistry, dapisCountReturndata!);
+      if (!dapisCountReturndata) {
```
Siegrift, line 70: This can't really happen. Use `!` to silence the warning instead.

src/update-feeds/update-feeds.ts
```diff
@@ -1,10 +1,13 @@
 import { go } from '@api3/promise-utils';
-import { ethers } from 'ethers';
-import { range, size } from 'lodash';
+import { BigNumber, ethers } from 'ethers';
```
Siegrift, line 2: nit: I'd prefer `ethers.BigNumber`

src/update-feeds/update-feeds.ts
```diff
@@ -41,6 +44,10 @@ export const startUpdateFeedLoops = async () => {
   );
 };
 
+export type ReadDapiWithIndexResponsesAndChainId = (ReadDapiWithIndexResponse & {
```
Siegrift, line 47: I don't think you need to do this. If I understand correctly, you only need to pass the `chainId` to the `processBatch` function, which you can do with as a new parameter.

src/update-feeds/update-feeds.ts
```diff
@@ -124,8 +138,84 @@ export const runUpdateFeed = async (providerName: string, chain: Chain, chainId:
   });
 };
 
-// eslint-disable-next-line @typescript-eslint/require-await
-export const processBatch = async (batch: ReadDapiWithIndexResponse[]) => {
+export const mergeUrls = (receivedUrls: UrlSet[], freshExistingUrls: UrlSet[]) =>
```
Siegrift, line 141: We don't want to merge the signed API URLs but replace them.

src/update-feeds/update-feeds.ts
```diff
@@ -124,8 +138,84 @@ export const runUpdateFeed = async (providerName: string, chain: Chain, chainId:
...
+  Object.values(groupBy([...receivedUrls, ...freshExistingUrls], 'url')).flatMap(
+    (group) => sortBy(group, 'lastReceivedMs').pop()!
+  );
+
+export const updateDynamicState = (batch: ReadDapiWithIndexResponsesAndChainId) => {
+  batch.map((item) =>
```
Siegrift, line 147: nit: Use just one `updateState` and and use a for loop inside it:
```ts
for (const dapi of batch) {...}
```

src/update-feeds/update-feeds.ts
```diff
@@ -124,8 +138,84 @@ export const runUpdateFeed = async (providerName: string, chain: Chain, chainId:
...
 };
 
-// eslint-disable-next-line @typescript-eslint/require-await
-export const processBatch = async (batch: ReadDapiWithIndexResponse[]) => {
+export const mergeUrls = (receivedUrls: UrlSet[], freshExistingUrls: UrlSet[]) =>
+  Object.values(groupBy([...receivedUrls, ...freshExistingUrls], 'url')).flatMap(
+    (group) => sortBy(group, 'lastReceivedMs').pop()!
+  );
+
+export const updateDynamicState = (batch: ReadDapiWithIndexResponsesAndChainId) => {
```
Siegrift, line 146: Why not just inline this?

src/update-feeds/update-feeds.ts
```diff
@@ -124,8 +138,84 @@ export const runUpdateFeed = async (providerName: string, chain: Chain, chainId:
...
+        dataFeedValues: { ...cachedDapiResponse?.dataFeedValues, [item.chainId]: item.dataFeedValue },
+        updateParameters: { ...cachedDapiResponse?.updateParameters, [item.chainId]: item.updateParameters },
+      };
+    })
+  );
+};
+
+export const getFeedsToUpdate = (batch: ReadDapiWithIndexResponsesAndChainId) =>
+  batch
+    .map((dapiResponse) => {
```
Siegrift, line 176: This returns 3 possible values:
1. dapi response + shouldUpdate=false
2. dapi response + signed data
3. false

It would be simpler if you do a map where you return the signed data (which can be undefined) for each data feed and then use `filter` to filter out those that we don't want to update.

src/update-feeds/update-feeds.ts
```diff
@@ -124,8 +138,84 @@ export const runUpdateFeed = async (providerName: string, chain: Chain, chainId:
...
+export const processBatch = async (batch: ReadDapiWithIndexResponsesAndChainId) => {
   logger.debug('Processing batch of active dAPIs', { batch });
-  // TODO: Implement.
+
+  // Start by merging the dynamic state with the state
+  updateDynamicState(batch);
+
+  const feedsToUpdate = getFeedsToUpdate(batch);
+
+  return Promise.allSettled(chunk(feedsToUpdate, FEEDS_TO_UPDATE_CHUNK_SIZE).map(async (feed) => updateFeeds(feed)));
```
Siegrift, line 220: Remove `FEEDS_TO_UPDATE_CHUNK_SIZE`. We agreed not to chunk these anymore (the whole dAPIs batch will be updated at once).

src/update-feeds/update-feeds.test.ts
```diff
@@ -1,18 +1,30 @@
...
 
 import {
   generateMockDapiDataRegistry,
   generateReadDapiWithIndexResponse,
 } from '../../test/fixtures/dapi-data-registry';
+import { generateTestConfig } from '../../test/fixtures/mock-config';
 import { allowPartial } from '../../test/utils';
 import type { DapiDataRegistry } from '../../typechain-types';
 import type { Chain } from '../config/schema';
+// import * as gasPriceModule from '../gas-price/gas-price';
```
Siegrift, line 12: Remove?

src/update-feeds/update-feeds.test.ts
```diff
@@ -180,6 +217,10 @@ describe(runUpdateFeed.name, () => {
       '123'
     );
 
+    expect(getStateSpy).toHaveBeenCalledTimes(4);
+    // TODO what has this fn been replaced with
```
Siegrift, line 221: Mhh, what do you mean?

src/update-feeds/dapi-data-registry.ts
```diff
@@ -24,6 +26,31 @@ export const decodeDapisCountResponse = (dapiDataRegistry: DapiDataRegistry, dap
 
 export type DapisCountResponse = ReturnType<typeof decodeDapisCountResponse>;
 
+export const decodeDataFeed = (dataFeed: string): DecodedDataFeed => {
+  if (dataFeed.length === 130) {
```
Siegrift, line 30: I think it would be nice to add a permalink to the contract to explain this encoding/decoding beahviour.

src/update-feeds/dapi-data-registry.ts
```diff
@@ -24,6 +26,31 @@ export const decodeDapisCountResponse = (dapiDataRegistry: DapiDataRegistry, dap
...
+    // (64 * 2) - 2
```
Siegrift, line 31:
```suggestion
    // (64 * 2) + 2
```

and it would be nice to explain why *2 and also +2

src/update-feeds/dapi-data-registry.ts
```diff
@@ -45,7 +76,7 @@ export const decodeReadDapiWithIndexResponse = (
       heartbeatInterval,
     },
     dataFeedValue: { value, timestamp },
-    dataFeed,
+    dataFeed: decodedDataFeed,
```
Siegrift, line 79: I think we could do a `decodedDataFeed: decodedDataFeed` to make this more explicit.

src/state/state.ts
```diff
@@ -2,19 +2,42 @@ import type { BigNumber } from 'ethers';
...
 import type { Config } from '../config/schema';
-import type { LocalSignedData, AirnodeAddress, TemplateId } from '../types';
+import type { chainId, DApiName, DecodedDataFeed, DataFeedId, SignedData } from '../types';
 
 interface GasState {
   gasPrices: { price: BigNumber; timestampMs: number }[];
   sponsorLastUpdateTimestampMs: Record<string, number>;
 }
 
+export interface DataFeedOnChainValue {
```
Siegrift, line 12: I would just call it `DataFeedValue`

src/state/state.ts
```diff
@@ -2,19 +2,42 @@ import type { BigNumber } from 'ethers';
...
+  value: BigNumber;
+  timestamp: number; // in seconds
```
Siegrift, line 14:
```suggestion
  timestamp: number;
```

We already have a strong convention that timestamp is in seconds. Whenever there is ms timestamp (except maybe in tests) there should be a suffix `Ms` in the variable name

src/state/state.ts
```diff
@@ -2,19 +2,42 @@ import type { BigNumber } from 'ethers';
...
+}
+
+export interface DApi {
```
Siegrift, line 17: The `DApi` feels to generic. I think we could be more explicit and call it `DapiState`

src/state/state.ts
```diff
@@ -31,6 +54,10 @@ export const getState = (): State => {
 
 export const setState = (newState: State) => {
   state = newState;
+
+  if (!state.dapis) {
```
Siegrift, line 58: Why?

src/gas-price/gas-price.ts
```diff
@@ -70,7 +70,14 @@ export const clearSponsorLastUpdateTimestampMs = (
   sponsorWalletAddress: string
 ) =>
   updateState((draft) => {
-    delete draft.gasPriceStore[chainId]![providerName]!.sponsorLastUpdateTimestampMs[sponsorWalletAddress];
+    const sponsorLastUpdateTimestampMs = get(
```
Siegrift, line 73: Get the value manually. You lose all typesafety when using `get` from Lodash.

src/update-feeds/update-feeds.ts
```diff
@@ -124,8 +138,84 @@ export const runUpdateFeed = async (providerName: string, chain: Chain, chainId:
...
+    updateState((draft) => {
+      const receivedUrls = item.signedApiUrls.flatMap((url) =>
+        item.dataFeed.dataFeeds.flatMap((dataFeed) => ({
+          url: `${url}/${dataFeed.airnodeAddress}`,
+          lastReceivedMs: Date.now(),
+        }))
+      );
+
+      const freshExistingUrls = draft.signedApiUrlStore.filter(
+        (url) => Date.now() - url.lastReceivedMs > 1000 * 60 * SIGNED_URL_EXPIRY_IN_MINUTES
```
Siegrift, line 157: We have not agreed to do this freshness check.

src/utils.ts
aquarat: Thanks for this, this actually was a mistake @ duplicate.

aquarat: It actually wasn't the same code/implementation - this function encodes a datafeed object as bytes - the whole object, not just the beacon ID.

src/update-feeds/update-feeds.ts
aquarat: Yes, it was based on when I thought (and implemented) a scenario where we retained locally defined URLs.

aquarat:
Please check now @Siegrift

src/utils.ts
Siegrift: > It actually wasn't the same code/implementation - this function encodes a datafeed object as bytes - the whole object, not just the beacon ID.

🤦  You're right.

src/signed-api-fetch/data-fetcher.ts
```diff
@@ -64,14 +63,8 @@ export const runDataFetcher = async () => {
...
 
-  const urls = uniq(
-    Object.keys(config.chains).flatMap((chainId) =>
-      Object.entries(signedApiUrlStore[chainId]!).flatMap(([airnodeAddress, baseUrl]) => `${baseUrl}/${airnodeAddress}`)
-    )
-  );
-
   return Promise.allSettled(
-    urls.map(async (url) =>
+    signedApiUrlStore.map(async (url) =>
```
Siegrift, line 67: Mhh, are these URLs unique per provider? The previous implementation looked correct to me.

src/state/state.ts
```diff
@@ -2,32 +2,44 @@ import type { BigNumber } from 'ethers';
...
 let state: State | undefined;
 
-export const getState = (): State => {
-  if (!state) {
-    throw new Error('State is undefined.');
-  }
-
-  return state;
-};
+export const getState = (): State => state!;
```
Siegrift, line 42: Why change this? It seems quite dengarous to me.

src/update-feeds/update-feeds.test.ts
```diff
@@ -167,6 +171,24 @@ describe(runUpdateFeed.name, () => {
     jest.spyOn(logger, 'debug');
     jest.spyOn(logger, 'error');
 
+    const testConfig = generateTestConfig();
+    jest.spyOn(stateModule, 'getState').mockReturnValue({
```
Siegrift, line 175: You can use the `allowPartial<State>` utility for slight advantages.

src/update-feeds/update-feeds.ts
```diff
@@ -103,9 +105,9 @@ export const runUpdateFeed = async (providerName: string, chain: Chain, chainId:
           await dapiDataRegistry.callStatic.tryMulticall(readDapiWithIndexCalls)
         );
 
-        const decodedBatch = returndata.map((returndata) =>
-          decodeReadDapiWithIndexResponse(dapiDataRegistry, returndata)
-        );
+        const decodedBatch = returndata.map((returndata) => ({
+          ...decodeReadDapiWithIndexResponse(dapiDataRegistry, returndata),
+        }));
```
Siegrift, line 110:
```suggestion
        const decodedBatch = returndata.map((returndata) => 
          decodeReadDapiWithIndexResponse(dapiDataRegistry, returndata),
        );
```

src/update-feeds/update-feeds.ts
```diff
@@ -124,8 +128,60 @@ export const runUpdateFeed = async (providerName: string, chain: Chain, chainId:
...
   logger.debug('Processing batch of active dAPIs', { batch });
-  // TODO: Implement.
+
+  updateState((draft) => {
+    for (const dapi of batch) {
+      const receivedUrls = dapi.signedApiUrls.flatMap((url) =>
+        dapi.decodedDataFeed.beacons.flatMap((dataFeed) => `${url}/${dataFeed.airnodeAddress}`)
+      );
+
+      draft.signedApiUrlStore = receivedUrls.flat();
```
Siegrift, line 174: These need to be indexed by provider. I think there is some confusion here.

We agreed on previous call was to store the latest URLs reported by each provider (these are to be replaced each time a batch is processed). When we call the signed APIs, we use all signed API URLs and for each beacon we save the latest signed data (based off the timestamp).

Siegrift (changes requested):

src/signed-api-fetch/data-fetcher.ts
aquarat: The previous implementation wasn't per provider, I've made these per provider now 👌

src/state/state.ts
aquarat: I just thought it'd make it a bit cleaner. The state being undefined should be impossible (there won't be any config) - but it doesn't bother me, I've reverted the change :)

src/update-feeds/update-feeds.ts
aquarat: Done. In retrospect an obvious requirement given that we replace URLs.

src/update-feeds/update-feeds.ts
```diff
@@ -171,7 +169,7 @@ export const processBatch = async (batch: ReadDapiWithIndexResponse[], chainId:
         dapi.decodedDataFeed.beacons.flatMap((dataFeed) => `${url}/${dataFeed.airnodeAddress}`)
       );
 
-      draft.signedApiUrlStore = receivedUrls.flat();
+      draft.signedApiUrlStore[providerName] = receivedUrls.flat();
```
Siegrift, line 172: We actually need to index by chainId as well, because it's possible to have two "infura" providers on different chains.

Siegrift (changes requested):
Probably the last comment 🙏

src/update-feeds/update-feeds.ts
aquarat: Done :D 

It is keyed as `signedApiUrlStore: Record<ChainId, Record<Provider, string[]>>` - which has to be unique.

Siegrift (approved):
👍 LGTM

PR #68, opened 2023-11-03
Use getAirseekerRecommendedGasPrice

### LINKED ISSUE #48: Use gas oracle for each dAPI in batch that needs an update
Blocked by https://github.com/api3dao/airseeker-v2/issues/47

After https://github.com/api3dao/airseeker-v2/issues/47 is done, we need to create and submit the update transactions. For that we need to calculate the gas price for each dAPI (taking into account that the transaction can be retry).

vponline:
Closes #48

### DIFF
--- a/src/update-feeds/update-feeds.ts
+++ b/src/update-feeds/update-feeds.ts
@@ -145,6 +145,8 @@ export const getFeedsToUpdate = (batch: ReadDapiWithIndexResponse[]) =>
       const offChainTimestamp = Number.parseInt(signedData?.timestamp ?? '0', 10);
       const deviationThreshold = updateParameters.deviationThresholdInPercentage;
 
+      // TODO clear last update timestamps if an update is not needed
+
       return checkUpdateConditions(
         dataFeedValue.value,
         dataFeedValue.timestamp,
--- a/src/gas-price/gas-price.test.ts
+++ b/src/gas-price/gas-price.test.ts
@@ -19,6 +19,10 @@ import {
 const chainId = '31337';
 const providerName = 'localhost';
 const rpcUrl = 'http://127.0.0.1:8545/';
+const provider = new ethers.providers.StaticJsonRpcProvider(rpcUrl, {
+  chainId: Number.parseInt(chainId, 10),
+  name: chainId,
+});
 const gasSettings = {
   recommendedGasPriceMultiplier: 1.5,
   sanitizationSamplingWindow: 15,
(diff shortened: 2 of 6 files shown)


src/update-feeds/update-feeds.ts
```diff
@@ -168,12 +169,13 @@ export const updateDynamicState = (batch: ReadDapiWithIndexResponsesAndChainId)
   );
 };
 
-export const getFeedsToUpdate = (batch: ReadDapiWithIndexResponsesAndChainId) =>
+export const getFeedsToUpdate = (batch: ReadDapiWithIndexResponsesAndChainId, sponsorWalletMnemonic: string) =>
   batch.map((dapiResponse) => {
     const signedData = getStoreDataPoint(dapiResponse.dataFeed.dataFeedId);
+    const sponsorWallet = ethers.Wallet.fromMnemonic(sponsorWalletMnemonic);
```
Siegrift, line 175: Why do we need to pass `sponsorWalletMnemonic` in this function? It seems it's just forwarded.

src/update-feeds/update-feeds.ts
```diff
@@ -193,27 +195,61 @@ export const getFeedsToUpdate = (batch: ReadDapiWithIndexResponsesAndChainId) =>
...
+export const updateFeeds = async (chains: Chains, batch: ReturnType<typeof getFeedsToUpdate>) => {
   // TODO implement
   // batch, execute
+
+  // Prepare to send transactions with each provider on each chain
+  const promisedTransactions = batch.flatMap((item) => {
+    const { chainId, sponsorWallet } = item;
+    const { providers, gasSettings } = chains[chainId]!;
+
+    Object.entries(providers).map(async ([providerName, { url }]) => {
```
Siegrift, line 211: This is not what we want to do. Each `runUpdateFeed` loop is run separately per provider. What we need to do is to pass the provider (or probably even better the contracts that are connected to the provider).

Siegrift (changes requested):

src/update-feeds/update-feeds.ts
vponline: Sorry I should have added a comment on this line, it's a placeholder. The mnemonic is needed to derive the datafeed specific wallet, right? I thought we might as well do it in this function so that the derived address can then be used to both clear update timestamps or send transactions.

Siegrift: But to derive the sponsor wallet we need the mnemonic (not the wallet) so this is not relevant. I still think it would be better to just pass the mnemonic to the relevant function that needs it instead of forwarding it.

vponline: I mean that we'd derive the wallet here, we need the derived wallet (address) to both clear update timestamps for feeds that don't need to be updated as well as to submit transactions for feeds that need to be updated, right?

Or do you mean that it would be ok to derive the wallet multiple times so we'd do it separately while clearing update timestamps as well as when submitting txs?

Siegrift: I mean we could just pass the `sponsorWalletMnemonic` (or get it from state) in `updateFeeds` function and derive the wallet there. It would tehn be used to both clear the timestamps and submit the update transactions.

vponline: Ah I think I get what you mean, then we just wouldn't filter feeds based on `shouldUpdate` in `processBatch`.

Siegrift (approved):
👍 LGTM

PR #69, opened 2023-11-03
Create and submit update transactions

Siegrift:
Closes https://github.com/api3dao/airseeker-v2/issues/36

Ideally merged after https://github.com/api3dao/airseeker-v2/pull/65 so I can rebase and use the decoded `dataFeed` value received from the contract.

## Rationalle

The logic is based of Airseeker v1. There could be some [optimisations](https://github.com/api3dao/airseeker-v2/issues/67) done, but I am not doing them just yet. 

There are some subtle differences, most notably in the sponsor wallet derivation and the general architecture (how we process the dAPIs in batches).


src/update-feeds/update-transactions.ts
```diff
@@ -0,0 +1,169 @@
...
+
+        if (!goUpdate.success) {
+          logger.error(`Failed to update a dAPI`, goUpdate.error);
+        }
+      });
+    })
+  );
+};
+
+export const estimateMulticallGasLimit = async (
```
Siegrift, line 95: Fully based on Airseeker v1 implementation. Later we might do https://github.com/api3dao/airseeker-v2/issues/67 but that needs a discussion and agreement first.

test/e2e/update-feeds.feature.ts
```diff
@@ -12,3 +17,59 @@ it('reads blockchain data correctly', async () => {
...
+        },
+        {
+          beaconId: krakenBtcBeacon.beaconId,
+          signedData: krakenBtcSignedData,
+        },
+      ],
+    },
+  ]);
+
+  expect(logger.debug).toHaveBeenCalledWith('Successfully updated dAPI');
```
Siegrift, line 74: We could also check the on-chain values, but I think this is enough.

src/update-feeds/update-transactions.ts
aquarat: Hey, AFAIK in airseeker v1 a configured gas limit is used (gas for calls is not estimated). 
Separately, manager-multisig often has to do multicalls and the gas is sometimes incorrectly estimated (often repeatedly for the same transaction). It's just something we should probably keep in mind; if we do estimates will we have a config parameter for adding headroom?

aquarat (commented):
(LGTM currently 🚀 ) - One comment about gas estimates

src/update-feeds/update-transactions.ts
Siegrift: > if we do estimates will we have a config parameter for adding headroom?

There is 10% headroom added when we do estimate of the `multicall`, so I guess that covers the need for `tryMulticall` and also adds this headroom.

aquarat: It's possible I made a mistake, thanks for clarifying :)
The current Airseeker implementation works well, so it must be fine :)

aquarat (approved):
LGTM 🔥 🚀

PR #71, opened 2023-11-06
Persist derived sponsor wallets

### LINKED ISSUE #70: Persist sponsor wallet private key
Deriving the sponsor wallet from HD node is performance intensive and unnecessary. Once we derive a sponsor wallet for a particular dAPI we should cache the derived wallet (or private key).

This can be done after the implementation is a bit more stable.

aquarat:
This PR adds persistence for derived sponsor wallets.

There is a considerable performance impact resulting from the process of deriving a sponsor wallet, so it's important to store the result for later re-use.

Closes #70


src/update-feeds/update-transactions.ts
```diff
@@ -135,6 +141,13 @@ export const deriveSponsorWallet = (sponsorWalletMnemonic: string, dapiName: str
   );
   logger.debug('Derived sponsor wallet', { sponsorAddress, sponsorWalletAddress: sponsorWallet.address });
 
+  updateState((draft) => {
+    draft.derivedSponsorWallets = {
+      ...draft.derivedSponsorWallets,
+      [dapiName]: sponsorWallet.privateKey,
+    };
+  });
```
vponline, line 149: 🕵️ could also:
```suggestion
  updateState((draft) => {
    draft.derivedSponsorWallets[dapiNake] = sponsorWallet.privateKey
  });
```

vponline (approved):
Nice 🚀

src/update-feeds/update-transactions.ts
aquarat: The current implementation is a cheat because `draft.derivedSponsorWallets` may be undefined. The spread approach creates the object if it doesn't exist.

vponline: ah right! Feel free to ignore then 😄

PR #73, opened 2023-11-07
Upgrade contract

Siegrift:
Closes https://github.com/api3dao/airseeker-v2/issues/66

The test [contract deploymemt in DapiDataRegistry tests](https://github.com/api3dao/dapi-management/blob/bf91e16be975cb99a1179c724a194cd6ab2c2e23/test/DapiDataRegistry.sol.js#L59) also deploys API3Fallback but we don't need that here (we just need to add dapis).


aquarat (approved):
LGTM 💪

PR #75, opened 2023-11-07
Deep beacon set update check

### LINKED ISSUE #67: Check which beacons of a beacon set need to be updated
See the [slack thread](https://api3workspace.slack.com/archives/C05TQPT7PNJ/p1697633154329109?thread_ts=1697564794.558719&cid=C05TQPT7PNJ) and a bit older [thread](https://api3workspace.slack.com/archives/C037NLX1R8T/p1690541486061689).

The old Airseeker implementation always attempts to update all beacons and uses `tryMulticall` to avoid issues with some updates reverting. This could be prevented by checking which beacons can be updated and updating only those.

Specifically, we could:
1. Get all on chain values of the data feed beacons
2. Filter beacons for which we have more up to date signed data (these could be updated)
3. Check whether updating beacons above would trigger the beacon set update (reuse on-chain value for non-updated beacons)
4. Do not use dummy data multicall estimation (there is no reason for multi call to revert apart from RPC provider failure).
5. Maybe not use the fixed 2_000_000 gas limit in case multicall estimation fails?
6. Use `multicall` instead of `tryMulticall` to do dAPI updates

aquarat:
This PR:
- Starts by executing the existing shallow beacon set check based on the datafeed's value returned from the chain during the batch call.
- For all feeds appearing to need an update it when reads the latest values of the children of those feeds and does two timestamp checks to confirm the feeds are indeed updatable.
- Child feeds that are not updatable have their on chain values used in a second/follow-up threshold assessment. If during that check the calcualted values don't exceed the deviation threshold the feeds are not updated. 
- Lastly, updatable child feeds are propagated to the updateFeeds function such that those feeds are not updated during the batch.

This PR also fixes a bug that PR #84 addresses, so will result in some merge conflicts. 

Closes #67

### DIFF
--- a/src/update-feeds/update-feeds.ts
+++ b/src/update-feeds/update-feeds.ts
@@ -1,6 +1,6 @@
 import { go } from '@api3/promise-utils';
 import { ethers } from 'ethers';
-import { chunk, range, size } from 'lodash';
+import { chunk, range, size, uniqBy } from 'lodash';
 
 import { calculateMedian, checkUpdateConditions } from '../condition-check';
 import type { Chain } from '../config/schema';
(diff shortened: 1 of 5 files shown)


src/update-feeds/update-feeds.ts
```diff
@@ -162,7 +164,154 @@ export const updateFeeds = async (_batch: ReturnType<typeof getFeedsToUpdate>, _
   // batch, execute
 };
 
-export const processBatch = async (batch: ReadDapiWithIndexResponse[], providerName: Provider, chainId: string) => {
+export const deepCheckFeeds = async (
```
Siegrift, line 167: You should modify the `getFeedsToUpdate` because it needs to determine which beacons are to be updated. This information should be propagated to `updateFeeds` function which should only update those beacons.

src/update-feeds/update-feeds.ts
```diff
@@ -162,7 +164,154 @@ export const updateFeeds = async (_batch: ReturnType<typeof getFeedsToUpdate>, _
...
+) => {
+  const { config } = getState();
+  const chain = config.chains[chainId]!;
+  const { providers, contracts } = chain;
+
+  const provider = new ethers.providers.StaticJsonRpcProvider(providers[providerName]);
+  const server = getApi3ServerV1(contracts.Api3ServerV1, provider);
+  const voidSigner = new ethers.VoidSigner(ethers.constants.AddressZero, provider);
+
+  // We surely have to chunk these? they'll get big.
```
Siegrift, line 180: The only reason the update can fail is the timestamp check https://github.com/api3dao/airnode-protocol-v1/blob/fa95f043ce4b50e843e407b96f7ae3edcf899c32/contracts/api3-server-v1/DataFeedServer.sol#L121

I think we should just get all the beacon values of this beacon set (using `readDataFeedWithId`). These are very simple so they can be multicalled together for the whole batch (I verified this by multicalling 10K `readDataFeedWithId` on Infura and the limits we are talking here are way smaller).

Siegrift (changes requested):

src/update-feeds/update-feeds.ts
Siegrift: Actually the `dataFeeds` function doesn't make the `require` call so it's even "cheaper" so we might call that.

aquarat:
Closing as rebasing was far too complex.

PR #81, opened 2023-11-08
Add clearing update timestamp for feeds with no update

### LINKED ISSUE #72: Clear last update timestamp for a dApi if update is not needed
Clear the last update timestamp for a dApi if an update is not needed for it

vponline:
Closes #72

### DIFF
--- a/src/update-feeds/update-feeds.ts
+++ b/src/update-feeds/update-feeds.ts
@@ -188,7 +188,7 @@ export const processBatch = async (batch: ReadDapiWithIndexResponse[], providerN
 
   const feeds = getFeedsToUpdate(batch);
 
-  // CLear last update timestamps for feeds that don't need an update
+  // Clear last update timestamps for feeds that don't need an update
   for (const feed of feeds.filter((feed) => !feed.shouldUpdate)) {
     clearSponsorLastUpdateTimestampMs(
       chainId,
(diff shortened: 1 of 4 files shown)


src/update-feeds/update-feeds.ts
```diff
@@ -186,5 +186,19 @@ export const processBatch = async (batch: ReadDapiWithIndexResponse[], providerN
     }
   });
 
-  return updateFeeds(getFeedsToUpdate(batch), chainId);
+  const feeds = getFeedsToUpdate(batch);
+
+  // CLear last update timestamps for feeds that don't need an update
```
Siegrift, line 191:
```suggestion
  // Clear last update timestamps for feeds that don't need an update
```

src/update-feeds/update-feeds.ts
```diff
@@ -186,5 +186,19 @@ export const processBatch = async (batch: ReadDapiWithIndexResponse[], providerN
...
+  for (const feed of feeds.filter((feed) => !feed.shouldUpdate)) {
```
Siegrift, line 192: nit: I would filter the ones that don't need to be updated based on the missing data feed IDs from the feeds to be updated (e.g. something along these lines `batch.filter(dapi => !feeds.includes(dapi.dataFeedId))`).

But I don't mind your approach either.

Siegrift (approved):
👍 LGTM

PR #84, opened 2023-11-09
Resolve TODOs

Siegrift:
Closes https://github.com/api3dao/airseeker-v2/issues/77

## Rationale

Each TODO is fixed in a separate commit for easier reviews. I recommend reviewing commit by commit.

I updated the diagram and exported the white background version. I don't remember the issues we had with it, but it seems correctly rendered to me. Let me know if I should change it back to black.

I also realized that the logic for determining updating beacon sets was wrong (we need to get data for constituent beacons and then use median for checks). I've updated the logic and created https://github.com/api3dao/airseeker-v2/issues/83.


src/update-feeds/update-transactions.ts
```diff
@@ -75,7 +75,6 @@ export const updateFeeds = async (
           );
 
           logger.debug('Deriving sponsor wallet');
-          // TODO: These wallets could be persisted as a performance optimization.
```
Siegrift, line 78: Done.

src/signed-api-fetch/data-fetcher.ts
```diff
@@ -32,7 +32,6 @@ const callSignedDataApi = async (url: string): Promise<SignedData[]> => {
         url,
         headers: {
           Accept: 'application/json',
-          // TODO add API key?
```
Siegrift, line 35: Signed APIs don't have authentication. There is an [issue](https://github.com/api3dao/signed-api/issues/22) created for that, but not planned in the short term.

src/config/schema.ts
```diff
@@ -22,7 +22,7 @@ export type Provider = z.infer<typeof providerSchema>;
 export const optionalContractsSchema = z
   .object({
     Api3ServerV1: evmAddressSchema.optional(),
-    DapiDataRegistry: evmAddressSchema, // TODO: Make optional and load from "airnode-protocol-v1" or some other location and document it accordingly.
```
Siegrift, line 25: https://github.com/api3dao/airseeker-v2/issues/82

src/types.ts
```diff
@@ -8,8 +8,7 @@ export type DapiName = string;
 export type PrivateKey = string;
 export type DataFeedId = EvmId;
 export type ChainId = string;
-export type DApiName = string;
-export type Provider = string;
+export type ProviderName = string;
```
Siegrift, line 11: The `Provider` makes me feel it's something from ethers.

src/update-feeds/update-transactions.ts
```diff
@@ -141,21 +141,16 @@ export const estimateMulticallGasLimit = async (
   return ethers.BigNumber.from(2_000_000);
 };
 
-export const deriveSponsorWallet = (sponsorWalletMnemonic: string, dapiName: string) => {
+export const getDerivedSponsorWallet = (sponsorWalletMnemonic: string, dapiName: string) => {
```
Siegrift, line 148: Had to split this function because it is stateful (checks the cache for private keys), but we want to use the `deriveSponsorWallet` as an utility in e2e test contract setup.

src/update-feeds/update-feeds.ts
```diff
@@ -128,42 +130,66 @@ export const runUpdateFeed = async (providerName: Provider, chain: Chain, chainI
...
+  return decodedBeaconValue;
+};
 
-    return {
-      ...dapiResponse,
-      signedData,
-      shouldUpdate,
-    };
-  });
+export const getFeedsToUpdate = (batch: ReadDapiWithIndexResponse[]): UpdateableDapi[] =>
```
Siegrift, line 149: @vponline I removed the `shouldUpdate` flag because I want this function to return the signed data for each beacon (so that we can use that data to update them) but if there is no data for a beacon there is nothing to return. I disliked making a new type that would have signed data optional so I just filtered the invalid values here directly.

src/types.ts
aquarat: It feels better as `ProviderName` 👍

aquarat:
> I also realized that the logic for determining updating beacon sets was wrong (we need to get data for constituent beacons and then use median for checks). I've updated the logic and created https://github.com/api3dao/airseeker-v2/issues/83.

ah yes, I separately realised that yesterday and updated it in my branch, whoops 🙈

aquarat (approved):
LGTM 🔥 
I didn't see any particular issues, all looks good. I'll keep these changes in mind for the inevitable merge conflict.

src/state/state.ts
```diff
@@ -41,8 +32,8 @@ export interface State {
   gasPriceStore: Record<string, Record<string, GasState>>;
   derivedSponsorWallets: Record<DapiName, PrivateKey>;
   signedApiStore: Record<DataFeedId, SignedData>;
-  signedApiUrlStore: Record<ChainId, Record<Provider, string[]>>;
-  dapis: Record<DApiName, DapiState>;
+  signedApiUrlStore: Record<ChainId, Record<ProviderName, string[]>>;
```
vponline, line 35: Should we also give a distinct name to the `SignedApiUrl` type here?
```suggestion
  signedApiUrlStore: Record<ChainId, Record<ProviderName, SignedApiUrl[]>>;
```

src/update-feeds/dapi-data-registry.ts
```diff
@@ -32,21 +32,21 @@ export const decodeDataFeed = (dataFeed: string): DecodedDataFeed => {
...
     const templateId = templateIds[idx] as string;
-    const dataFeedId = deriveBeaconId(airnodeAddress, templateId)!;
+    const beaconId = deriveBeaconId(airnodeAddress, templateId)!;
 
-    return { dataFeedId, airnodeAddress, templateId };
+    return { beaconId, airnodeAddress, templateId };
   });
 
-  const dataFeedId = deriveBeaconSetId(dataFeeds.map((df) => df.dataFeedId))!;
+  const dataFeedId = deriveBeaconSetId(beacons.map((df) => df.beaconId))!;
```
vponline, line 47: maybe 😄 
```suggestion
  const dataFeedId = deriveBeaconSetId(beacons.map((b) => b.beaconId))!;
```

vponline (approved):
LGTM, noticed a couple minor points 🔥

src/update-feeds/update-feeds.ts
```diff
@@ -186,19 +216,27 @@ export const processBatch = async (batch: ReadDapiWithIndexResponse[], providerN
...
 
-  const feeds = getFeedsToUpdate(batch);
+  const feedsToUpdate = getFeedsToUpdate(batch);
+  const dapiNamesToUpdate = new Set(feedsToUpdate.map((feed) => feed.dapiInfo.dapiName));
 
   // Clear last update timestamps for feeds that don't need an update
-  for (const feed of feeds.filter((feed) => !feed.shouldUpdate)) {
+  for (const feed of batch) {
+    if (dapiNamesToUpdate.has(feed.dapiName)) continue;
+
```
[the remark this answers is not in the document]
Siegrift: I don't mind either.

src/update-feeds/update-feeds.ts
```diff
@@ -120,50 +126,74 @@ export const runUpdateFeed = async (providerName: Provider, chain: Chain, chainI
...
+        updateableBeacons: zip(dapiInfo.decodedDataFeed.beacons, beaconsSignedData).map(([beacon, signedData]) => ({
+          signedData: signedData!,
+          beaconId: beacon!.beaconId,
+        })),
+      };
+    })
+    .filter((updateableDapi): updateableDapi is UpdateableDapi => updateableDapi !== null);
 
-export const processBatch = async (batch: ReadDapiWithIndexResponse[], providerName: Provider, chainId: string) => {
+export const processBatch = async (batch: ReadDapiWithIndexResponse[], providerName: ProviderName, chainId: string) => {
```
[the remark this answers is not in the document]
Siegrift: Although I am starting to wonder about the benefit of these specific string types. It feels good in typescript `Record` but in interfaces and function parameters you have the variable name which is often enough to deduce the full context.

PR #88, opened 2023-11-12
Make Airseeker run locally

Siegrift:
Closes https://github.com/api3dao/airseeker-v2/issues/78
Relates to https://github.com/api3dao/airseeker-v2/issues/63

## Rationale

When I tried running Airseeker docker locally I run into couple of issues, which I was fixing iteratively. The changes are quite small so I didn't bother splitting it into multiple commits. Specifically, I needed to do the following changes:

1. Fix the docker image. I've done it similarly to https://github.com/api3dao/signed-api/blob/0bad6fc8dd6aaffaa12cf099ab6bbf7c98d487c8/Dockerfile#L17
2. Fix `signedDataFetchInterval` example values (they were in ms).
3. Moved Typechain types to `src/typechain-types`. This is to have flattened folder structure in `/dist` which is important for Docker entrypoint.
4. Implemented `src/index.ts` to start the Airseeker loops.
5. Implement https://github.com/api3dao/airseeker-v2/issues/78
6. Fix minor bugs, change spammy/confusing log messages and add TODOs (which should be resolved before merging).


docker/Dockerfile
```diff
@@ -1,66 +1,28 @@
...
-
-# Build preparation
-FROM environment AS preparation
-
-WORKDIR ${buildDir}
-
-RUN apk add --update --no-cache git
-
-# Install pnpm
+# Install pnpm globally (needs root privilidges).
```
aquarat, line 6:
```suggestion
# Install pnpm globally (needs root privileges).
```

docker/Dockerfile
```diff
@@ -1,66 +1,28 @@
...
-# Create Airseeker user
-RUN addgroup -S ${name} && \
-    adduser -h ${appDir} -s /bin/false -S -D -H -G ${name} ${name} && \
-    chown -R ${name} ${appDir}
-
-USER ${name}
-
-CMD ["node", "dist/index.js"]
\ No newline at end of file
+# Create a new user and group to run the application with reduced priviledges.
```
aquarat, line 9:
```suggestion
# Create a new user and group to run the application with reduced privileges.
```

docker/Dockerfile
```diff
@@ -1,66 +1,28 @@
...
+RUN addgroup -S airseeker && \
+    adduser -h /app -s /bin/false -S -D -H -G airseeker airseeker && \
```
aquarat, line 11: It's probably unnecessary to use a separate user for this in the container? I think most services use root inside docker containers or pass-through the user and group ids from the launch shell. The root-in-container process's root status shouldn't translate into root on the host machine. Feel free to educate me 😆

src/update-feeds/update-feeds.ts
```diff
@@ -40,6 +40,9 @@ export const startUpdateFeedLoops = async () => {
 
       for (const providerName of Object.keys(providers)) {
         logger.debug(`Starting update feed loop`, { chainId, providerName });
+        // Run the update feed loop manually for the first time, because setInterval first waits for the given period of
+        // time.
+        void runUpdateFeed(providerName, chain, chainId);
```
aquarat, line 45: It seems there are no timeouts currently, so it's possible for these calls to overlap potentially quite a lot?

aquarat (approved):
LGTM, minor comments/observations 🔥

src/gas-price/gas-price.ts
```diff
@@ -47,6 +47,7 @@ export const clearExpiredStoreGasPrices = (chainId: string, providerName: string
  * @param providerName
  * @param sponsorWalletAddress
  */
+// TODO: Why is this unused?
```
vponline, line 50: I don't think the `update-transactions` file existed and we probably forgot to use it 😄

This function should be used directly after submitting an update transaction, but only if it's the first time submitting a particular update (so not reset on when scaling is applied)..

src/gas-price/gas-price.ts
```diff
@@ -171,6 +172,7 @@ export const getAirseekerRecommendedGasPrice = async (
     scalingWindow,
     maxScalingMultiplier,
   } = gasSettings;
+  initializeGasStore(chainId, providerName); // TODO: This could be moved to a better place to avoid calling it multiple times (in different functions).
```
vponline, line 175: It could probably just be removed, We'd never call this function without first having collected gas prices, which means that the `gasPriceStore` would already be initialized.

test/fixtures/mock-config.ts
```diff
@@ -32,6 +32,7 @@ export const generateTestConfig = (): Config => ({
   deviationThresholdCoefficient: 1,
 });
 
+// TODO: Do we need this function? Can't we use the one from production code?
```
vponline, line 35: I assume you mean `setInitialState`, which looks like we could 😄

vponline (approved):

docker/Dockerfile
Siegrift: I blame https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker

docker/Dockerfile
Siegrift: It one of the best practices. You can read a bit more in:
- https://stackoverflow.com/a/68190847
- https://snyk.io/blog/10-best-practices-to-containerize-nodejs-web-applications-with-docker/
- (and many more)

Running the service as a non root user is quite simple to do so it's better to stick with the best practice.

src/gas-price/gas-price.ts
Siegrift: Not really, when we first call this function we have no gas prices stored.

src/signed-api-fetch/data-fetcher.ts
```diff
@@ -89,7 +89,6 @@ export const runDataFetcher = async () => {
             localDataStore.setStoreDataPoint(element);
           }
         },
-        // TODO: What to do about timeout and retries? This is the only place that does some handling.
```
Siegrift: https://github.com/api3dao/airseeker-v2/issues/91

As a note. This is the only axios call at the moment. The thing that should change is the timeout. It should not be a static value but instead depend on the `signedDataFetchInterval`.

src/update-feeds/update-feeds.ts
```diff
@@ -42,17 +42,17 @@ export const startUpdateFeedLoops = async () => {
...
         await sleep(staggerTime);
       }
     })
   );
 };
 
-export const runUpdateFeed = async (providerName: ProviderName, chain: Chain, chainId: ChainId) => {
-  await logger.runWithContext({ chainId, providerName, coordinatorTimestampMs: Date.now().toString() }, async () => {
+export const runUpdateFeeds = async (providerName: ProviderName, chain: Chain, chainId: ChainId) => {
+  await logger.runWithContext({ chainId, providerName, updateFeedsCoordinatorId: Date.now().toString() }, async () => {
```
Siegrift, line 58: I renamed the function after I renamed the "coordinator ID". It actually updates many feeds so plural makes more sense. 

It creates slight awkwardness with `startUpdateFeedsLoops` but the name is actually right. There are many loops which update many feeds.

src/signed-api-fetch/data-fetcher.ts
```diff
@@ -42,12 +57,12 @@ const callSignedDataApi = async (url: string): Promise<SignedData[]> => {
     }
   );
 
-  // TODO: Pass the response error body (if available). See OEV repo for details.
-  if (!result.success) {
-    throw new Error([`HTTP call failed: `, url, result.error].join('\n'));
```
Siegrift, line 45: In general, we should not rethrow errors because it:
1. Messes up with the stack trace (using Error.cause mitigates this though)
2. The point of catching the errors is likely to process it somehow. If we process an error and re-throw it will likely be handled twice which is not good from design perspective.

src/gas-price/gas-price.ts
```diff
@@ -10,9 +10,7 @@ export const initializeGasStore = (chainId: string, providerName: string) =>
       draft.gasPriceStore[chainId] = {};
     }
 
-    if (!draft.gasPriceStore[chainId]![providerName]) {
-      draft.gasPriceStore[chainId]![providerName] = { gasPrices: [], sponsorLastUpdateTimestampMs: {} };
-    }
+    draft.gasPriceStore[chainId]![providerName] = { gasPrices: [], sponsorLastUpdateTimestampMs: {} };
```
Siegrift, line 13: This will now always to the initialisation for a particular provider, other provider data for the chain is untouched. 

Maybe it could also reset the gas price store for all providers and all chains? (It's not really important because the initialisation is now called once per provider).

docker/Dockerfile
aquarat: Thanks for this 👌

PR #90, opened 2023-11-13
Config signed urls

### LINKED ISSUE #76: Implement signed APIs in config
There should be a config option that allows specifying global signed APIs (these are called with all the signed APIs reported by RPC providers across all chains).

vponline:
Closes #76


src/signed-api-fetch/data-fetcher.test.ts
```diff
@@ -1,31 +1,29 @@
...
 
-import { runDataFetcher, stopDataFetcher } from './data-fetcher';
+import * as dataFetcherModule from './data-fetcher';
 
 const mockedAxios = axios as jest.MockedFunction<typeof axios>;
 jest.mock('axios');
 
 describe('data fetcher', () => {
   beforeEach(() => {
+    jest.resetAllMocks();
```
Siegrift, line 14: I was puzzled by this, because we use `mockRestore: true` in the config. This should call restoreAllMocks before each test, which should clear the mocks, reset them and also restore the original values. See:
- https://jestjs.io/docs/configuration#restoremocks-boolean
- https://jestjs.io/docs/mock-function-api#mockfnmockreset

This makes mockRestore the superset of mock reset and mock clear. However, there is a catch. Restoring mocks only works for spyOn calls and jest.mock is special and restore mocks doesn't work for it https://stackoverflow.com/a/56512217.

Which means we should add `mockReset: true` to the config as well. That option works even for jest.mock calls.

Siegrift (approved):
👍 LGTM after fixing the minor comments.

PR #92, opened 2023-11-13
Only store signed data for active dapis

### LINKED ISSUE #79: Only store fresh signed data
Currently, we do not store active dAPIs in state which means that we need to store the signed data for every beacon even if there are only a few active beacons among the active dAPIs.

Another side effect of this is that once we fetch a data for a beacon, we persist it forever.

aquarat:
Closes #79 

This PR filters incoming signed data to only store signed data for active beacons.
This PR also purges data in the signed data store for which there are no active dApis.

I'm not sure if putting `getBeaconsForActiveDapis` in State is a good idea 🤷


Siegrift:
I've unassigned myself from reviewer list. Per the discussion on the call the implementtion needs to change to time based cleanup after 1d. 

Feel free to request review again when it's fixed.

aquarat:
CLosing as replaced by https://github.com/api3dao/airseeker-v2/pull/100 (due to a combination of conflicts and change in strategy)

PR #95, opened 2023-11-15
Check which beacons of a beacon set need to be updated

### LINKED ISSUE #67: Check which beacons of a beacon set need to be updated

aquarat:
This PR:
- Starts by executing the existing shallow beacon set check based on the beacon set's value returned from the chain during the batch call.
- For all feeds appearing to need an update it reads the latest values of the children of those feeds and does two timestamp checks to confirm the feeds are indeed updatable.
- Child feeds that are not updatable have their on-chain values used in a second/follow-up threshold assessment. If during that check the calculated values don't exceed the deviation threshold the feeds are not updated. 
- Lastly, updatable child feeds are propagated to the updateFeeds function such that those feeds are not updated during the batch.

Closes #67 

Replaces https://github.com/api3dao/airseeker-v2/pull/75 (refactoring this was difficult/it was easier to start from main and re-implement)


src/update-feeds/update-feeds.ts
```diff
@@ -152,52 +151,34 @@ export const decodeBeaconValue = (encodedBeaconValue: string) => {
   return decodedBeaconValue;
 };
 
-export const getFeedsToUpdate = (
+// https://github.com/api3dao/airnode-protocol-v1/blob/fa95f043ce4b50e843e407b96f7ae3edcf899c32/contracts/api3-server-v1/DataFeedServer.sol#L132
+export const encodeBeaconValue = (numericValue: string) => {
```
Siegrift, line 155: It seems this is only relevant for `check-feeds.ts`. Can you move it there?

src/update-feeds/update-transactions.ts
```diff
@@ -10,22 +10,22 @@ import { deriveSponsorWallet } from '../utils';
...
+  UpdatableBeacons: UpdatableBeacon[];
 }
 
 export const updateFeeds = async (
   chainId: ChainId,
   providerName: ProviderName,
   provider: ethers.providers.StaticJsonRpcProvider,
   api3ServerV1: Api3ServerV1,
-  updateableDapis: UpdateableDapi[]
+  UpdatableDapis: UpdatableDapi[]
```
Siegrift, line 28:
```suggestion
  updatableDapis: UpdatableDapi[]
```

(not capitalized).

src/update-feeds/update-feeds.ts
```diff
@@ -152,52 +151,34 @@ export const decodeBeaconValue = (encodedBeaconValue: string) => {
...
-          dapiInfo.dataFeedValue.timestamp,
-          newBeaconSetValue,
-          newBeaconSetTimestamp,
-          dapiInfo.updateParameters.heartbeatInterval,
-          adjustedDeviationThresholdCoefficient
-        )
-      ) {
-        return null;
-      }
+  deviationThresholdCoefficient: DeviationThresholdCoefficient,
```
Siegrift, line 163: Mhh, why not just use `number`? I undertand why use the named types for chain ID, provider name, etc... But I don't see the benefit here.

src/update-feeds/check-feeds.ts
```diff
@@ -0,0 +1,198 @@
...
+      };
+    })
+    .filter((UpdatableDapi): UpdatableDapi is UpdatableDapi => UpdatableDapi !== null);
+
+interface OnChainValue {
+  beaconId: string;
+  onChainValue: { timestamp: ethers.BigNumber; value: ethers.BigNumber };
+}
+
+export const callAndParseMulticall = async (
```
vponline, line 69: This is cool, checking the multicall results is always a lot of lines 😅

src/update-feeds/check-feeds.ts
```diff
@@ -0,0 +1,198 @@
...
+  batch: string[],
+  providerName: ProviderName,
+  chainId: ChainId
+): Promise<OnChainValue[]> => {
+  const { config } = getState();
+  const chain = config.chains[chainId]!;
+  const { providers, contracts } = chain;
+
+  const provider = new ethers.providers.StaticJsonRpcProvider(providers[providerName]);
```
vponline, line 78: We could probably pass the `provider` in as an input, but not a big deal either way 

I'd say don't do that here, I'll do it in my PR when I have to rebase on this PR since I have similar changes done there

src/update-feeds/check-feeds.ts
```diff
@@ -0,0 +1,198 @@
...
+    .filter((onChainValue): onChainValue is OnChainValue => onChainValue !== null);
+};
+
+export const deeplyCheckDapis = (
+  batch: ReturnType<typeof shallowCheckFeeds>,
+  deviationThresholdCoefficient: DeviationThresholdCoefficient,
+  onChainValues: Awaited<ReturnType<typeof callAndParseMulticall>>
+): UpdatableDapi[] =>
+  batch
+    .map(({ dapiInfo, UpdatableBeacons }) => {
```
vponline, line 130: might need to lowercase?
```suggestion
    .map(({ dapiInfo, updatableBeacons }) => {
```

vponline (approved):
LGTM 🔥

src/update-feeds/check-feeds.test.ts
```diff
@@ -0,0 +1,265 @@
...
+          '0x000a': { timestamp: '100', encodedValue: encodeBeaconValue('200') },
+          '0x000b': { timestamp: '150', encodedValue: encodeBeaconValue('250') },
+          '0x000c': { timestamp: '200', encodedValue: encodeBeaconValue('300') },
+        });
+      });
+
+      const batch = allowPartial<ReadDapiWithIndexResponse[]>([
+        {
+          updateParameters: { deviationThresholdInPercentage: ethers.BigNumber.from(2) },
+          dataFeedValue: { value: ethers.BigNumber.from(150), timestamp: 150 },
```
Siegrift, line 56: nit: It would be nice to set `timestamp: 300` so its obvious it fails because of the timestamp check.

src/update-feeds/update-feeds.ts
```diff
@@ -152,52 +151,34 @@ export const decodeBeaconValue = (encodedBeaconValue: string) => {
...
+  providerName: ProviderName,
+  chainId: ChainId
+): Promise<UpdatableDapi[]> => {
+  const shallowCheckedDapis = shallowCheckFeeds(batch, deviationThresholdCoefficient);
```
Siegrift, line 167: Doing a shallow check filtering first and afterwards doing the deep filter will not work. Consider this case:
1. Say we have all data for a beacon set update but the aggregated value does not update the feed, but some of our signed data is outdated.
2. The `shallowCheckFeeds` will omit this beacon set.
3. However, it is still possible that the beacon set needs an update (when we replace the outdated signed data values with on-chain ones).

This needs to be implemented in one function, roughly like this:
1. Get all the beacons for all dAPIs in the batch and multicall their values (these should be de-duped by beacon ID).
2. Check update conditions for each dAPI separately.
3. For each beacon use the more up to date option (on-chain, off-chain) and do the aggregation and update check.
4. If the dAPI needs an update, only update the beacons for which you used the off-chain values.

Siegrift (changes requested):
Also, as part of this issue we should remove the [dummy gas estimation logic ](https://github.com/api3dao/airseeker-v2/blob/6dfa280a105a68b384939a9047d005603c314cdd/src/update-feeds/update-transactions.ts#L138) because the update transaction can't revert (unless of some rare race condition, but for that there is the generous static gas limit).

src/update-feeds/update-feeds.ts
aquarat: I think this is now implemented as required 👌 
LMK if I need to handle a multicall failure with a fallback 🤷 🙏

src/update-feeds/update-feeds.test.ts
```diff
@@ -262,8 +263,8 @@ describe(updateFeedsModule.runUpdateFeeds.name, () => {
   });
 });
 
-describe(updateFeedsModule.getFeedsToUpdate.name, () => {
-  it('applies deviationThresholdCoefficient from config', () => {
+describe('tODO', () => {
```
Siegrift, line 266: I guess this is forgotten.

src/update-feeds/check-feeds.ts
```diff
@@ -0,0 +1,164 @@
...
+import { getState } from '../state';
+import type { ChainId, ProviderName } from '../types';
+import { multiplyBigNumber } from '../utils';
+
+import { getApi3ServerV1 } from './api3-server-v1';
+import type { ReadDapiWithIndexResponse } from './dapi-data-registry';
+import { decodeBeaconValue } from './update-feeds';
+import type { UpdatableDapi } from './update-transactions';
+
+export const checkFeeds = async (
```
Siegrift, line 16: Why change the name from `getUpdatableFeeds`?

src/update-feeds/check-feeds.ts
```diff
@@ -0,0 +1,164 @@
...
+      .filter((dapi) => dapi.updatableBeacons.length > 0)
+  );
+};
+
+interface OnChainValue {
+  beaconId: string;
+  onChainValue: { timestamp: ethers.BigNumber; value: ethers.BigNumber };
+}
+
+export const callAndParseMulticall = async (
```
Siegrift, line 111: This is specific for reading dataFeeds. Can we rename it to something more explicit? E.g. `multicallBeaconValues`

src/update-feeds/check-feeds.ts
```diff
@@ -0,0 +1,164 @@
...
+  batch: string[],
```
Siegrift, line 112: nit: 
```suggestion
  batch: BeaconId[],
```

src/update-feeds/check-feeds.ts
```diff
@@ -0,0 +1,164 @@
...
+        updatableBeacons: dapiInfo.decodedDataFeed.beacons
+          .filter(({ localDataIsStale }) => !localDataIsStale)
+          .map(({ beaconId, airnodeAddress, signedData, templateId }) => ({
+            beaconId,
+            airnodeAddress,
+            signedData,
+            templateId,
+          })),
+      }))
+      // Finally, filter out dapis that cannot be updated (eg. if we don't have signed data)
```
Siegrift, line 101:
```suggestion
      // Finally, filter out dapis that cannot be updated (e.g. if we don't have signed data)
```

Siegrift: But can this case actually happen? I think if there are no updatable beacons then the dAPI will already be excluded in the first filter.

src/update-feeds/check-feeds.ts
```diff
@@ -0,0 +1,164 @@
...
+        );
+      })
+      // Transform the batch and exclude beacons that cannot be updated
+      .map((dapiInfo) => ({
+        dapiInfo,
+        updatableBeacons: dapiInfo.decodedDataFeed.beacons
+          .filter(({ localDataIsStale }) => !localDataIsStale)
+          .map(({ beaconId, airnodeAddress, signedData, templateId }) => ({
+            beaconId,
+            airnodeAddress,
```
Siegrift, line 96: `UpdatableBeacon` needs just `beaconId` and `signedData`.

src/update-feeds/check-feeds.ts
```diff
@@ -0,0 +1,164 @@
...
+              encodedValue: ethers.constants.HashZero,
+              templateId: ethers.constants.HashZero,
+              signature: ethers.constants.HashZero,
+            };
+            const offChainValue = {
+              timestamp: ethers.BigNumber.from(signedData.timestamp),
+              value: decodeBeaconValue(signedData.encodedValue)!,
+            };
+
+            const localDataIsStale =
```
Siegrift, line 52: nit: `localDataIsStale` -> `localDataIsValid` because of the future check as well. But good point about that.

Siegrift (approved):
👍 LGTM (as usual feel free to ignore the nits).

src/update-feeds/update-feeds.test.ts
aquarat: Aaaargh, sorry!

src/update-feeds/check-feeds.ts
aquarat: ah, it doesn't bother me, I think this was the name of the original parent function. I'll revert it to `getUpdatableFeeds` 👌

src/update-feeds/check-feeds.ts
aquarat: mmm, I suspect it could maybe happen if the data on chain is updated between the application getting signed data and reading the on chain values? That would be a very rare scenario though. I can remove this check.

src/update-feeds/check-feeds.ts
aquarat: 😆 indeed 🙈

PR #98, opened 2023-11-15
Refactor timeouts

### LINKED ISSUE #91: Handle timeouts and properly cleanup memory
See discussion in https://github.com/api3dao/airseeker-v2/pull/88#discussion_r1390934475

Axios [does not have any timeout by default](https://stackoverflow.com/a/57995963). Ethers porovider has a 2 minute default timeout (but we can maybe set it explicitly). These are important to properly cleanup resources.

The go utils timeout can be applied as necessary. E.g. ensure that some loop calls do not overlap. This is tricky to do correctly because it's impossible to cancel running an already started function.

vponline:
Closes #91


src/signed-api-fetch/data-fetcher.ts
```diff
@@ -38,23 +38,21 @@ export const stopDataFetcher = () => {
...
-        url,
-        headers: {
-          Accept: 'application/json',
-        },
-      }),
-    {
-      attemptTimeoutMs: HTTP_SIGNED_DATA_API_ATTEMPT_TIMEOUT,
-      totalTimeoutMs: HTTP_SIGNED_DATA_API_ATTEMPT_TIMEOUT + HTTP_SIGNED_DATA_API_HEADROOM / 2,
-      retries: 0,
-    }
```
vponline, line 57: I don't think we need these anymore, right? The `axios` request timeout should be enough to resolve the `go` promise. Let me know if I'm missing something 😅

src/signed-api-fetch/data-fetcher.ts
```diff
@@ -38,23 +38,21 @@ export const stopDataFetcher = () => {
...
+export const callSignedDataApi = async (
+  url: string,
+  signedDataFetchIntervalMs: number
+): Promise<SignedData[] | null> => {
+  const goAxiosCall = await go<Promise<AxiosResponse>, AxiosError>(async () =>
+    axios({
+      method: 'get',
+      timeout: signedDataFetchIntervalMs * HTTP_SIGNED_DATA_API_HEADROOM_MULTIPLIER,
```
aquarat, line 50: Does this need `Math.ceil` or equivalent? I feel like it could get weird with float rounding errors 🤷

aquarat (approved):
LGTM 🔥

src/update-feeds/update-feeds.ts
```diff
@@ -57,32 +58,39 @@ export const startUpdateFeedsLoops = async () => {
...
+        const firstBatch = readDapiWithIndexCallsReturndata
+          // Because the dapisCount is not known during the multicall, we may ask for non-existent dAPIs. These should be filtered out.
+          .slice(0, dapisCount)
+          .map((dapiReturndata) => ({ ...decodeReadDapiWithIndexResponse(dapiDataRegistry, dapiReturndata), chainId }));
+        return {
+          firstBatch,
+          dapisCount,
+        };
+      },
+      { totalTimeoutMs: dataFeedUpdateIntervalMs, attemptTimeoutMs: dataFeedUpdateIntervalMs }
```
Siegrift, line 92: You only need to set the `totalTimeoutMs`.

However, when we call `runUpdateFeeds` and fetching the first batch takes `dataFeedUpdateIntervalMs` we should return and not process other batches (because they would overlap with another round of `runUpdateFeeds`).

Siegrift (changes requested):
The timeout logic in `runUpdateFeeds` needs to change imo.

src/update-feeds/update-transactions.ts
```diff
@@ -40,75 +40,82 @@ export const updateFeeds = async (
...
-          logger.debug('Updating dAPI', { gasPrice: goGasPrice.data, gasLimit });
-          return (
-            api3ServerV1
-              // When we add the sponsor wallet (signer) without connecting it to the provider, the provider of the
-              // contract will be set to "null". We need to connect the sponsor wallet to the provider of the contract.
-              .connect(sponsorWallet.connect(api3ServerV1.provider))
-              .tryMulticall(dataFeedUpdateCalldatas, { gasPrice, gasLimit })
-          );
-        });
+        const goUpdate = await go(
```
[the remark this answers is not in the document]
vponline: The formatting changed because I added timeouts to the `goUpdate` function and `goGasPrice`

src/signed-api-fetch/data-fetcher.ts
vponline: There can't by any harm to round it and it's pretty much always a good idea 😅

src/update-feeds/update-transactions.ts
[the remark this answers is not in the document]
Siegrift: Oh, right.

src/update-feeds/update-feeds.ts
```diff
@@ -89,12 +89,17 @@ export const runUpdateFeeds = async (providerName: ProviderName, chain: Chain, c
...
         };
       },
-      { totalTimeoutMs: dataFeedUpdateIntervalMs, attemptTimeoutMs: dataFeedUpdateIntervalMs }
+      { totalTimeoutMs: dataFeedUpdateIntervalMs }
     );
     if (!goFirstBatch.success) {
       logger.error(`Failed to get first active dAPIs batch`, goFirstBatch.error);
       return;
     }
+    if (Date.now() >= firstBatchStartTime + dataFeedUpdateIntervalMs) {
```
Siegrift, line 98: It's still not bulletproof, but I am not sure we can actually implement this in a bulletproof way.

Siegrift (approved):
👍 LGTM

PR #100, opened 2023-11-16
Purge old signed data

aquarat:
This PR adds code which:
- Prevents Airseeker from storing signed data with a timestamp older than 24 hours
- Purges data older than 24 hours from the store after every API fetch

Closes [#92](https://github.com/api3dao/airseeker-v2/issues/79)
Replaces https://github.com/api3dao/airseeker-v2/pull/92


src/signed-data-store/signed-data-store.test.ts
```diff
@@ -75,4 +77,27 @@ describe('datastore', () => {
...
+      draft.signedApiStore['0x001'] = allowPartial<SignedData>({
+        timestamp: (baseTime / 1000 - 23 * 60 * 60).toString(),
+      });
+    });
+
+    purgeOldSignedData();
+
+    expect(Object.values(getState().signedApiStore)).toStrictEqual([
+      {
+        timestamp: '1700043430',
```
Siegrift, line 99: nit: `(baseTime / 1000 - 23 * 60 * 60).toString()` ?

maybe save those values as `purgableTimestamp` and `nonPurgableTimestamp`

Siegrift (approved):
👍 LGTM.

I wonder if we should call as part of the signed data fetch loop (which is quite high frequency), but I think it's fine and we can optimize later if needed.

aquarat:
> 👍 LGTM.
> 
> I wonder if we should call as part of the signed data fetch loop (which is quite high frequency), but I think it's fine and we can optimize later if needed.

Yeah, it had crossed my mind... it could be on a separate interval instead, easy to change. The performance impact of doing it on the fetcher interval should be very very minor.

PR #103, opened 2023-11-20
Update feeds tests

### LINKED ISSUE #83: Write unit tests for feeds that need to be updated
The function `getUpdatableFeeds` should be tested to determine that the logic for determining feeds to update is correct.

vponline:
Closes #83


Siegrift (approved):
👍 Nice, LGTM.

Issue #96, opened 2023-11-15
Investigate CI e2e test issue


Siegrift:
See: https://github.com/api3dao/airseeker-v2/actions/runs/6863104836/job/18662177317

But it has happened a few times as well. Re-running doesn't help (at least from my attempts) but amend + force-pushing does. I was not able to reproduce it locally, nor find any relevant information.

We may also remove/merge/update the test if this keeps happening and we are unable to find the reason.

aquarat:
Interesting, the error message can be reproduced by killing Hardhat while the test is running.

vponline:
I saw this today: https://github.com/NomicFoundation/hardhat/issues/4601

PR #104, opened 2023-11-20
Update hardhat

vponline:
Updating all hardhat related dependencies just in case 
For #96


Siegrift:
I ran into some (unrelated) local development issues and tried updating HH in https://github.com/api3dao/airseeker-v2/pull/162 so merging that supercedes this PR.

Issue #85, opened 2023-11-09
Come up with strategy around multicall batch sizes


Siegrift:
Batch sizes are likely going to be different per provider as they can apply their own limits. We need to set the limit for multicalling active dAPIs.

See [this thread](https://api3workspace.slack.com/archives/C05TQPT7PNJ/p1699529274108669) for reference.

bbenligiray:
A similar concern is eth_getLogs limits and providers used in an Airnode config. We just set a blanket limit for the chain there and weed out the providers that don't work with it.

aquarat:
Could we maybe do a quick benchmark when Airseeker starts?

aquarat:
It was decided to hardcode batch sizes to 10.

PR #105, opened 2023-11-20
Possible batch sizing detection

aquarat:
An idea for automatically detecting the support batch sizes per provider.

Relates to #85


aquarat:
Closing as decided against this approach in a meeting.

PR #112, opened 2023-11-22
Add CloudFormation template

Siegrift:
Closes https://github.com/api3dao/airseeker-v2/issues/108

## Rationale

Pretty much the Pusher template. For now without the Loki integration, because I don't have proper access yet. It will be done separately and is tracked in https://github.com/api3dao/airseeker-v2/issues/111


Siegrift:
I'm merging because I've tested this by deploying it on AWS and it is running.

PR #119, opened 2023-11-28
Fix error handling and update transaction calldata

Siegrift:
Closes https://github.com/api3dao/airseeker-v2/issues/113

## Rationale

I implemented the error handling as the issue says. As I was changing the `multicallBeaconValues` and `getUpdatableFeeds` I decided to simplify it a bit by avoiding mutation in map/filter but I preserved the logic.

I also noticed that we are incorrectly creating the beacon set update calldata. We need to update the beacon set with all beacons (not just the ones that are updatable). I've added a few tests to guard for this.

### DIFF
--- a/src/condition-check/condition-check.ts
+++ b/src/condition-check/condition-check.ts
@@ -48,7 +48,6 @@ export const checkDeviationThresholdExceeded = (
  *
  * Update transaction with stale data would revert on chain, draining the sponsor wallet. See:
  * https://github.com/api3dao/airnode-protocol-v1/blob/dev/contracts/dapis/DataFeedServer.sol#L121
- * This can happen if the gateway or Airseeker is down and Airkeeper does the updates instead.
  */
 export const checkFulfillmentDataTimestamp = (onChainDataTimestamp: number, fulfillmentDataTimestamp: number) =>
   onChainDataTimestamp < fulfillmentDataTimestamp;
@@ -71,7 +70,6 @@ export const checkUpdateConditions = (
   const isFulfillmentDataFresh = checkFulfillmentDataTimestamp(onChainTimestamp, offChainTimestamp);
   if (!isFulfillmentDataFresh) {
     logger.warn(`Off-chain sample's timestamp is older than on-chain timestamp.`);
-
     return false;
   }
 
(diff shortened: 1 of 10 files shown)


src/condition-check/condition-check.ts
```diff
@@ -48,7 +48,6 @@ export const checkDeviationThresholdExceeded = (
  *
  * Update transaction with stale data would revert on chain, draining the sponsor wallet. See:
  * https://github.com/api3dao/airnode-protocol-v1/blob/dev/contracts/dapis/DataFeedServer.sol#L121
- * This can happen if the gateway or Airseeker is down and Airkeeper does the updates instead.
```
Siegrift, line 51: This part of the comment was outdated.

src/condition-check/condition-check.ts
```diff
@@ -82,12 +80,10 @@ export const checkUpdateConditions = (
     const shouldUpdate = checkDeviationThresholdExceeded(onChainValue, deviationThreshold, offChainValue);
     if (shouldUpdate) {
       logger.info(`Deviation exceeded.`);
-
       return true;
     }
   } else {
-    logger.info(`On-chain timestamp is older than the heartbeat interval.`);
-
+    logger.debug(`On-chain timestamp is older than the heartbeat interval.`);
```
Siegrift, line 86: The info level feels to high for this type of information.

src/update-feeds/update-feeds.ts
```diff
@@ -56,104 +56,133 @@ export const startUpdateFeedsLoops = async () => {
...
 
-    // Calculate the stagger time between the rest of the batches.
-    const batchesCount = Math.ceil(dapisCount / dataFeedBatchSize);
-    const staggerTime = batchesCount <= 1 ? 0 : (dataFeedUpdateInterval / batchesCount) * 1000;
+      const { firstBatch, dapisCount } = goFirstBatch.data;
+      if (dapisCount === 0) {
+        logger.warn(`No active dAPIs found`);
+        return;
+      }
+      // NOTE: We need to explicitly handle the .catch here because it's possible that the promise settles before it's
```
Siegrift, line 113: This gave me a hard time to debug, but its actually correct. There is a `sleep` call below. If the promise resolves sooner it is unhandled (because we await it only a few lines below). To address this issue we need to catch the error ourselves.

aquarat (approved):
LGTM, very neat, good to have the "bad path" tested 😅

PR #120, opened 2023-11-28
Fix warning message

Siegrift:
Closes https://github.com/api3dao/airseeker-v2/issues/114


aquarat (approved):
LGTM, I'll keep `.name` in mind for the future

Siegrift:
> LGTM, I'll keep .name in mind for the future

I think it's a neat trick and the benefit is that renaming refactor will also change the name of the test (describe block).

PR #133, opened 2023-11-28
Remove stopDataFetcher

Siegrift:
Closes https://github.com/api3dao/airseeker-v2/issues/129


aquarat (approved):
LGTM
If you remove the stop function you probably don't need the interval stored in state either, but again, super minor 🤷

Siegrift:
> If you remove the stop function you probably don't need the interval stored in state either, but again, super minor

Yeah, I removed that in some later PR but I could have done it here.

PR #134, opened 2023-11-28
Make explicit distinction between dapiName and decodedDapiName

Siegrift:
## Rationale

As I was doing updates in the codebase I got a bit lost in whether the `dapiName` is the encoded or decoded version. We were mixing that in tests (because it often didn't matter) and were using it also in logs (but we should always use the decode dAPI name for convenience).

I made the distinction by introducing `decodedDapiName` and introducing a convention that `dapiName` is the encoded version. I've updated the tests and logs to reflect this.

### DIFF
--- a/local-test-configuration/scripts/initialize-chain.ts
+++ b/local-test-configuration/scripts/initialize-chain.ts
@@ -312,9 +312,9 @@ export const deploy = async (funderWallet: ethers.Wallet, provider: ethers.provi
     tx = await dapiDataRegistry
       .connect(api3MarketContract)
       .addDapi(
-        dapiName,
-        beaconSetId,
-        sponsorWalletMnemonic,
+        dapiName!,
+        beaconSetId!,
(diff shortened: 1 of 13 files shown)


bbenligiray:
FWIW I was looking at https://github.com/api3dao/airseeker-v2/blob/21b13bd5b64f2ab52d5e02494fe923237a42c005/src/utils.ts#L32 today and was horrified because I thought `dapiName` was not encoded

Siegrift:
> and was horrified because I thought dapiName was not encoded because of the variable name/type

We are mimicking the types from the contract when possible and the [contract returns dapiName](https://github.com/api3dao/dapi-management/blob/d47515295a586c2db6d9f2ee34fb91c6ed4a902c/contracts/DapiDataRegistry.sol#L396) but yeah, the confusion is understandable.

aquarat (approved):
Great idea, LGTM 🔥

PR #142, opened 2023-11-29
Sponsor wallet warning

Siegrift:
Closes https://github.com/api3dao/airseeker-v2/issues/140


aquarat (approved):
LGTM 
Interesting, I guess it should have been  a hash of the name instead, but that would have been gas-expensive.

Siegrift:
> Interesting, I guess it should have been a hash of the name instead, but that would have been gas-expensive.

Mhh, what do you mean by gas-expensive? This is only needed to derive the sponsor wallet path. I actually like your suggestion by hashing the dapiName and using that for derivation. WDYT @bbenligiray ?

PR #153, opened 2023-12-04
WIP: Validate chain ID

Siegrift:
Closes https://github.com/api3dao/airseeker-v2/issues/148

## Rationale

As discussed in the issue, we want to log a warning and discard the provider if it reports a wrong chain ID (or there is an error getting chain ID). This check is done only once, at start-time.


Siegrift:
Superceded by https://github.com/api3dao/airseeker-v2/issues/154

PR #159, opened 2023-12-12
Use historical saved gas price when RPC gas price fails

### LINKED ISSUE #147: Use historical gas prices when RPC call for gas price fails
See: https://api3workspace.slack.com/archives/C05TQPT7PNJ/p1701327746032679

The exact details needs to be thought through. For example:
1. What defines a good enough set of historical prices?

Siegrift:
Closes https://github.com/api3dao/airseeker-v2/issues/147

## Rationale

I changed the `getRecommendedGasPrice` to reuse the latest saved gas price (if fresh) and made the function non-throwing and instead return `null` if there is no gas price to use.

### DIFF
--- a/src/gas-price/gas-price.test.ts
+++ b/src/gas-price/gas-price.test.ts
@@ -1,6 +1,6 @@
 import { ethers } from 'ethers';
 
-import { initializeState } from '../../test/fixtures/mock-config';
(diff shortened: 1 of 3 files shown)


aquarat (approved):
LGTM 👍

PR #162, opened 2023-12-12
Use dot at the end of log messages, update hardhat dependencies

### LINKED ISSUE #146: Improve consistency of log statements
All logs should end with dot for consistency.

Siegrift:
Closes https://github.com/api3dao/airseeker-v2/issues/146

## Rationale

I was running into wierd issues when building Typechain artifacts and I thought it's because of some wierd dependency issue, so I updated them (but it didn't help). I managed to fix that by clearing HH artifacts (running `pnpm run clean` helped).


aquarat (approved):
> so I updated them (but it didn't help). I managed to fix that by clearing HH artifacts (running `pnpm run clean` helped).

So adding '.' to the log messages does not relate to the issue you experienced? nor the dependency bumps? hah, just checking 😆 

Regardless, it LGTM 👍

Siegrift:
No, I wasn't able to push (without `--no-verfify`) so I spent some time investigating why... Since I updated the dependencies I didn't want to revert that work.

PR #166, opened 2023-12-18
Update docker image name in instructions

Siegrift:
As I was doing the Docker release I noticed that I forgot to change the container name when I copy pasted the instructions from signed API. I also removed suffix `-v2` from the images becuase we are using the `api3/airseeker` image.


aquarat (approved):
ah makes sense, LGTM 🔥

PR #168, opened 2023-12-31
Multicall chain info

### LINKED ISSUE #154: Multicall chain info with read calls, use tx count from the fetched block
See: https://api3workspace.slack.com/archives/C05S589E7B4/p1701700197041599

Siegrift:
Closes https://github.com/api3dao/airseeker-v2/issues/154

## Rationale

See the issue description, comments and related Slack discussions for full context. 

We fetch on-chain data in two places, 1) when fetching a batch of active data feeds and 2) when multi-calling beacon values for the batch. Both of these verify that chain ID matches the configuration and log a warning otherwise. During 1) we remember the block ID and use it to fetch the nonce (at that block) when making the update transaction. In 2) we fetch the values at `latest` the block number. See [this discussion](https://api3workspace.slack.com/archives/C05TQPT7PNJ/p1704718141612169?thread_ts=1704709566.544549&cid=C05TQPT7PNJ).

When making an update transaction we log the sponsor wallet address and a nonce of the update transaction.

I've also updated the `local-test-configuration` which was broken after https://github.com/api3dao/airseeker-v2/pull/165.

### DIFF
--- a/local-test-configuration/airnode-feed-1/airnode-feed.json
+++ b/local-test-configuration/airnode-feed-1/airnode-feed.json
@@ -36,12 +36,13 @@
   "signedApis": [
     {
       "name": "localhost",
-      "url": "http://${LOCALHOST_IP}:4001"
+      "url": "http://${LOCALHOST_IP}:4001",
+      "authToken": null
     }
   ],
   "ois": [
     {
-      "oisFormat": "2.2.0",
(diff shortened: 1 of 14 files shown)


aquarat (approved):
I can't offer you any feedback 😆😅, it's extremely neat ⭐ 👏

PR #171, opened 2024-01-01
Upgrade contracts3 update

bbenligiray:
I made two changes to AirseekerRegistry:
- I decided that the contract logic wasn't perfectly unambiguous and the data feed ID and dAPI name-based interfaces/storage have to be separated properly. Had to change one line of e2e test setup for this.
- https://github.com/api3dao/dapi-management/blob/8470f78f829c69f871fd1988217c47eab209f222/contracts/AirseekerRegistry.sol#L340 was weird so I had `activeDataFeed()` also return `dataFeedId`. The current airseeker-v2 implementation derives the data feed ID from `dataFeedDetails`, which is fine to keep as is.

### DIFF
--- a/src/update-feeds-loops/contracts.ts
+++ b/src/update-feeds-loops/contracts.ts
@@ -37,9 +37,9 @@ export const decodeDataFeedDetails = (dataFeed: string): DecodedDataFeed | null 
   // https://github.com/api3dao/dapi-management/blob/f3d39e4707c33c075a8f07aa8f8369f8dc07736f/contracts/AirseekerRegistry.sol#L209
   if (dataFeed === '0x') return null;
 
-  if (dataFeed.length === 130) {
-    // (64 [actual bytes] * 2[hex encoding] ) + 2 [for the '0x' preamble]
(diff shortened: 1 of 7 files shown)


src/update-feeds-loops/contracts.ts
```diff
@@ -73,6 +73,10 @@ export const decodeUpdateParameters = (updateParameters: string): DecodedUpdateP
     ['uint256', 'int224', 'uint256'],
     updateParameters
   );
+  // 2 characters for the '0x' preamble + 3 parameters, 32 * 2 hexadecimals for 32 bytes each
+  if (updateParameters.length !== 2 + 3 * (32 * 2)) {
```
bbenligiray, line 77: The user (AirseekerRegistry owner) can specify `updateParameters` of arbitrary length, which is why it is important to check this imo. I check it after the decoding attempt because the error message for it will probably be more descriptive.

Siegrift: Good point. We should definitely do this.

Siegrift (approved):

PR #178, opened 2024-01-30
Migrate from ethers v5 to ethers v6

### LINKED ISSUE #177: Upgrade to ethers-v6
Related to https://github.com/api3dao/airseeker-v2/issues/172

As part of migration to https://github.com/bbenligiray/api3-contracts for up-to-date version of AirseekerRegistry, we need to update to ethers-v6 because the repository provided (as of version 0.0.2) only factories for ethers-v6. Because the v5 version is no longer actively maintained, it makes sense to make a switch to v6.

Siegrift:
Closes https://github.com/api3dao/airseeker-v2/issues/177

## Rationale

I followed https://docs.ethers.org/v6/migrating/ and tried to advance by small steps (that's why so many commits). Usually going from file to file, understanding the TS issue and fixing it globally. There were a few gotchas worth mentioning:
1. Multiple things are not mentioned in the migration guide
2. Be extra careful when [deriving sponsor wallet](https://github.com/api3dao/airseeker-v2/blob/dfe6d14f540730c947d8fb4f222f91743f8f9053/src/utils.ts#L39). Especially if these are [ported to commons](https://github.com/api3dao/commons/blob/8b52aeecb082560c17223cc026d727d41240ac96/src/blockchain-utilities/derivation.ts#L55) (maybe as part of https://github.com/api3dao/commons/issues/29)
3. Working with Hardhat is considerably slower. Seems like an internal change in the underlying provider implementation.
4. https://github.com/api3dao/airseeker-v2/pull/178/commits/379e0d0ed5fdd8d944066a9b18962940df588733 fixes a breaking change with how ethers "ArrayObjects" work. In ethers v5 when you fetched an event or called a function, you received back for example:
```ts
Promise<[BigNumber, number] & {
    value: BigNumber;
    timestamp: number;
}>
```

when you did `Object.keys(suchResponse)` you got back `['0', '1', 'value', 'timestamp']`. In ethers v6, what you get is `['0', '1']` (the named keys are not own properties). This breaks utilities such as `omit`.

## Quality assurance

Apart from fixing all of the tests, I ran Airseeker locally for a few hours and it updated the feeds correctly so I think all is fine. I 

## TODO
- [x] Write PR description
- [x] Fix TODOs in code
- [x] Consider using BigInt literal (e.g. 123n) instead of BigInt constructor
- [x] Fix CI test
- [x] Run e2e local test flow
- [x] Ensure that `ethers.Wallet.fromPhrase` derives the same wallet as `ethers.Wallet.fromMnemonic` in ethers v5
- [x] Fix tests


contracts/airnode-protocol-v1-contracts.sol
Siegrift, line 1: We need to compile the contracts now because the package only exports ether v5 factories.

bbenligiray: https://github.com/api3dao/tasks/issues/153 would be a solution I guess

Siegrift: Yeah. It's not a big issue for Airseeker as it only impacts from which dependency it gets the contracts (or factories). But it's very good that you publish the contracts so that consumers can build them any way they want.

local-test-configuration/monitoring/index.html
```diff
@@ -741,52 +797,46 @@ <h2>Active data feeds</h2>
...
+      return arr.length % 2 === 0 ? (nums[mid - 1] + nums[mid]) / 2n : nums[mid];
     };
 
     const decodeBeaconValue = (encodedBeaconValue) => {
       // Solidity type(int224).min
-      const INT224_MIN = ethers.BigNumber.from(2).pow(ethers.BigNumber.from(223)).mul(ethers.BigNumber.from(-1));
+      const INT224_MIN = 2n ** 223n * -1n;
       // Solidity type(int224).max
-      const INT224_MAX = ethers.BigNumber.from(2).pow(ethers.BigNumber.from(223)).sub(ethers.BigNumber.from(1));
+      const INT224_MAX = 2n ** 223n - 1n;
```
aquarat, line 812: Very cool, I didn't realise there was a native BigInt or that it could be specified like this.

src/deviation-check/deviation-check.ts
```diff
@@ -1,54 +1,49 @@
...
-export const calculateDeviationPercentage = (initialValue: ethers.BigNumber, updatedValue: ethers.BigNumber) => {
-  const delta = updatedValue.sub(initialValue);
-  const absoluteDelta = delta.abs();
+export const calculateDeviationPercentage = (initialValue: bigint, updatedValue: bigint) => {
+  const delta = updatedValue - initialValue;
+  const absoluteDelta = abs(delta);
 
   // Avoid division by 0
-  const absoluteInitialValue = initialValue.isZero() ? ethers.BigNumber.from(1) : initialValue.abs();
+  const absoluteInitialValue = initialValue === 0n ? BigInt(1) : abs(initialValue);
```
aquarat: Why not `21_000n` instead of `BigInt`? (this question applies to any instance of a `BigInt(literal)`). Just curious, it's super minor/no right or wrong.

local-test-configuration/monitoring/index.html
Siegrift: Yeah, in v6 of ethers they switched the BigNumber library to native BigInt so we have the advantage of some of the built in operators.

aquarat (approved):
LGTM 👍 only a very very minor comment around BigInt literals.
This was good for me as I have a better idea now of the differences between ethers v5 and v6 🚀

local-test-configuration/monitoring/index.html
aquarat: I suspect it will be much more efficient for processing too.

src/deviation-check/deviation-check.ts
Siegrift: There is no difference and I prefer literal, but when I was doing the migration I basically just replaced all `ethers.BigNumber` with `BigInt` (and had to fix the issue that BigInt throws when the number is float). 

I didn't want to waste too much time on rewriting these, but I figured now it's quite easy with a regex so I've replaced them. Would love to see ESLint doing this automatically though.

aquarat: All good, I was just curious :)

PR #182, opened 2024-02-03
Self funded wallet derivation

### LINKED ISSUE #179: Support self-funded sponsor derivation
See: https://api3workspace.slack.com/archives/C05TQPT7PNJ/p1706521432528779.

Agreed on call that this is the best way to support self-funded dAPIs. Use the derivation that Nodary uses and add a setting in the config which chooses the right option. Don't forget to document it in the README.

Siegrift:
Closes https://github.com/api3dao/airseeker-v2/issues/179

## Rationale

We have not discussed the details (mostly naming) so I went ahead with the following:
1. Added `walletDerivationScheme` in the root of the `airseeker.json` config.
2. The supported values are `self-funded` and `managed`.
See the [README.md changes](https://github.com/api3dao/airseeker-v2/compare/read-feeds-values-from-chain...self-funded-wallet-derivation?expand=1#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5) for details.

The wallet derivation changes are [here](https://github.com/api3dao/airseeker-v2/compare/read-feeds-values-from-chain...self-funded-wallet-derivation?expand=1#diff-39b2554fd18da165b59a6351b1aafff3714e2a80c1435f2de9706355b4d32351) and the self-funded sponsor derivation takes into account the update parameters as referenced [here](https://api3workspace.slack.com/archives/C05TQPT7PNJ/p1706601083254309?thread_ts=1706521432.528779&cid=C05TQPT7PNJ).

## TODO

- [x] Mention this change in the spec.

### DIFF
--- a/README.md
+++ b/README.md
@@ -230,6 +230,7 @@ The derivation scheme used to derive sponsor wallets. The following options are 
 
 - `self-funded` - The sponsor wallet is derived from the hash of the encoded beacon ID together with the update
   parameters. This is the scheme that was originally used by Nodary for self-funded data feeds.
+  <!-- TODO: Change to "managed" -->
 - `managed-dapis` - Derives the wallet from the hash of the dAPI name (or data feed ID). This means the wallet
   derivation is agnostic to update parameters, and the same wallet is used when the dAPI is upgraded/downgraded.
 
--- a/README.md
+++ b/README.md
@@ -224,6 +224,15 @@ The fetch interval in seconds between retrievals of signed API data.
 
 A list of signed API URLs to call along with URLs fetched from the chain.
 
+#### `walletDerivationScheme`
+
+The derivation scheme used to derive sponsor wallets. The following options are available:
+
+- `self-funded` - The sponsor wallet is derived from the hash of the encoded beacon ID together with the update
+  parameters. This is the scheme that was originally used by Nodary for self-funded data feeds.
+- `managed-dapis` - Derives the wallet from the hash of the dAPI name (or data feed ID). This means the wallet
+  derivation is agnostic to update parameters, and the same wallet is used when the dAPI is upgraded/downgraded.
+
 ## Docker
 
 ### Build
--- a/README.md
+++ b/README.md
@@ -230,9 +230,8 @@ The derivation scheme used to derive sponsor wallets. The following options are 
 
 - `self-funded` - The sponsor wallet is derived from the hash of the encoded beacon ID together with the update
   parameters. This is the scheme that was originally used by Nodary for self-funded data feeds.
-  <!-- TODO: Change to "managed" -->
-- `managed-dapis` - Derives the wallet from the hash of the dAPI name (or data feed ID). This means the wallet
-  derivation is agnostic to update parameters, and the same wallet is used when the dAPI is upgraded/downgraded.
+- `managed` - Derives the wallet from the hash of the dAPI name (or data feed ID). This means the wallet derivation is
+  agnostic to update parameters, and the same wallet is used when the dAPI is upgraded/downgraded.
 
 ## Docker
 
--- a/README.md
+++ b/README.md
@@ -226,7 +226,11 @@ A list of signed API URLs to call along with URLs fetched from the chain.
 
 #### `walletDerivationScheme`
 
-The derivation scheme used to derive sponsor wallets. The following options are available:
+The derivation scheme configuration used to derive sponsor wallets.
+
+##### `type`
+
+The following options are available:
 
 - `self-funded` - The sponsor wallet is derived from the hash of the encoded beacon ID together with the update
   parameters. This is the scheme that was originally used by Nodary for self-funded data feeds.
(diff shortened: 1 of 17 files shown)


local-test-configuration/monitoring/index.html
```diff
@@ -858,6 +865,23 @@ <h2>Active data feeds</h2>
       return sponsorWallet;
     };
 
+    const deriveSponsorWallet = (
+      sponsorWalletMnemonic,
+      dapiNameOrDataFeedId,
+      updateParameters,
+      walletDerivationScheme
+    ) => {
+      // Derive the hash of the data feed using which the sponsor wallet is derived. For self-funded feeds it's more
```
aquarat: This is a bit strange, but it's super minor (especially seeing as it's part of the _monitoring_ for the test configuration).

src/utils.ts
```diff
@@ -43,6 +53,23 @@ export const deriveSponsorWallet = (sponsorWalletMnemonic: string, dapiNameOrDat
   return sponsorWallet;
 };
 
+export const deriveSponsorWallet = (
+  sponsorWalletMnemonic: string,
+  dapiNameOrDataFeedId: string,
+  updateParameters: string,
+  walletDerivationScheme: WalletDerivationScheme
+) => {
+  // Derive the hash of the data feed using which the sponsor wallet is derived. For self-funded feeds it's more
```
aquarat: ah and here, I'm guessing this was probably copied from somewhere.

aquarat (approved):
LGTM (again 🤣 ) 🚀

local-test-configuration/monitoring/index.html
```diff
@@ -858,6 +865,23 @@ <h2>Active data feeds</h2>
...
+      // suitable to derive the hash also from update parameters. This does not apply to mananaged feeds which want to be
+      // funded by the same wallet independently of the update parameters.
+      const sponsorAddressHash =
```
bbenligiray, line 879: This is the sponsor address, not the sponsor address hash, no?

README.md
```diff
@@ -224,6 +224,15 @@ The fetch interval in seconds between retrievals of signed API data.
 
 A list of signed API URLs to call along with URLs fetched from the chain.
 
+#### `walletDerivationScheme`
+
+The derivation scheme used to derive sponsor wallets. The following options are available:
```
bbenligiray: Doesn't feel future compatible with https://api3workspace.slack.com/archives/C05TQPT7PNJ/p1706712947591099?thread_ts=1706521432.528779&cid=C05TQPT7PNJ

Siegrift: Yeah, but does it really need to be? I don't think it's a big issue if the Airseeker config changes and I wanted to prioritise clarity for current implementation. That said, if you want it future proof, we might do and object right away:
- `{ type: "managed" }`
- `{ type: "self-funded" }`
- (In the future) `{ type: "fixed-wallet", "mnemonic": "..." }` (mnemonic may not be needed if we use the Airseeker one in which case we could get away with just union of strings).

local-test-configuration/monitoring/index.html
Siegrift: No, I've separated the hashing part into separate functions and one base function (`deriveSponsorWalletFromSponsorAddressHash`) which takes first 20 bytes from the hash and interprets that as sponsor address and derives the wallet.

src/utils.ts
Siegrift: Yeah. Basically, all the functions in the local monitoring are copied from the implementation (and maybe slightly updated) so there is just a single HTML file that can be opened in browser.

local-test-configuration/monitoring/index.html
Siegrift: Yeah, the comment is a bit confusing. I've tried rewording it to better reflect what's implemented. Let me know if it makes sense.

local-test-configuration/monitoring/index.html
bbenligiray: I mean the thing that you split into chunks and build the path out of is called the sponsor address, regardless if it's the address of an actual sponsor (such as [these](https://github.com/api3dao/operations-database/blob/48676e0e79114b5250f21cf7518100b0eb5e6645/src/scripts/contants.ts))

README.md
bbenligiray: Yeah, add the `{}`, it's definitely worth it

local-test-configuration/monitoring/index.html
Siegrift: Yeah, that's why I don't want to name it `sponsorAddress`. There are:
- `sponsorAddressHash` - Hash from which we get the build path - e.g. `ethers.keccak256(ethers.solidityPacked(['bytes32', 'bytes'], [dapiNameOrDataFeedId, updateParameters]))`
- `sponsorAddress` - `ethers.getAddress(sponsorAddressHash.slice(0, 42))`
- `sponsorWallet` - Derived from mnemonic and `sponsorAddress`

Not sure if we are not misunderstanding each other. LMK if I am missing something.

README.md
Siegrift: Done.

local-test-configuration/monitoring/index.html
bbenligiray: Yeah okay, I hadn't dug deep enough to find `sponsorAddress`

PR #193, opened 2024-02-28
Refactor the utilizing of signed API URLs

### LINKED ISSUE #190: Fix handling of signed API URLs
I encountered the following error when I deployed the airseeker-v2:

```
{"cause":{"address":"127.0.0.1","code":"ECONNREFUSED","errno":-111,"port":80,"syscall":"connect"},"code":"ECONNREFUSED","dataFetcherCoordinatorId":"1708953860436","level":"warn","message":"connect ECONNREFUSED 127.0.0.1:80","ms":"+0ms","name":"Error","stack":"Error: connect ECONNREFUSED 127.0.0.1:80\n    at AxiosError.from (/app/node_modules/.pnpm/axios@1.6.5/node_modules/axios/dist/node/axios.cjs:837:14)\n    at RedirectableRequest.handleRequestError (/app/node_modules/.pnpm/axios@1.6.5/node_modules/axios/dist/node/axios.cjs:3090:25)\n    at RedirectableRequest.emit (node:events:529:35)\n    at RedirectableRequest.emit (node:domain:489:12)\n    at eventHandlers.<computed> (/app/node_modules/.pnpm/follow-redirects@1.15.4_debug@4.3.4/node_modules/follow-redirects/index.js:37:24)\n    at ClientRequest.emit (node:events:517:28)\n    at ClientRequest.emit (node:domain:489:12)\n    at Socket.socketErrorListener (node:_http_client:501:9)\n    at Socket.emit (node:events:517:28)\n    at Socket.emit (node:domain:489:12)","timestamp":"2024-02-26T13:24:20.441Z"}
```

After investigating it, I found out that when [activeDataFeed](https://github.com/api3dao/contracts/blob/3276b9452a443d50c0f8fbe5a09343e66d7e37b5/contracts/api3-server-v1/AirseekerRegistry.sol#L360) function returns empty string for `signedApiUrls` (it's when `signedApiUrls` isn't set in the `AirseekerRegistry`), airseeker uses empty string as base URL and create a complete signed API URL (including airnode address) to make request, see [here](https://github.com/api3dao/airseeker-v2/blob/a99a98acb3a8235edb15aadf8c1274e77281c1a1/src/update-feeds-loops/update-feeds-loops.ts#L272). It creates URLs like `/airnodeAddress` and make request for them. 

Besides that, as I understand from [here](https://github.com/api3dao/airseeker-v2/blob/a99a98acb3a8235edb15aadf8c1274e77281c1a1/src/data-fetcher-loop/data-fetcher-loop.ts#L84), airseeker expects to see complete signed API URLs for each airnode in the config file like following:
```json
{
"signedApiUrls": [
"https://signed-api.stage.nodary.io/public/0x5791Fb78D4e37A9D0f0003199D1AE1A8C04C8d89", 
"https://signed-api.stage.nodary.io/public/0xc52EeA00154B4fF1EbbF8Ba39FDe37F1AC3B9Fd4", 
"https://signed-api.stage.nodary.io/public/0x8676eA8B6Ebe5b8FBbc25FF55192bADf39D7D61b"]
}
```
This setup complicates airseeker deployments because each time a new partner provider (airnode) is added, all airseekers must be redeployed. We only want to re-deploy airseekers when we add/remove chains.

My suggestion here:

- Behave signed-api URLs in configuration same with the ones on the contracts. So if we expect to read URL without airnode address in the contract, we should also expect to read URL without airnode address in the configuration file
- Detect empty URLs coming from the contract and don't include them into URLs to be called

bdrhn9:
Closes #190


src/update-feeds-loops/update-feeds-loops.ts
```diff
@@ -241,6 +241,15 @@ export const runUpdateFeeds = async (providerName: ProviderName, chain: Chain, c
         dataFeedUpdates,
         dataFeedUpdateFailures,
       });
+
+      // Update the state with the signed API URLs.
+      const signedApiUrls = uniq(
+        processedBatches.reduce((acc, batch) => (batch ? [...acc, ...batch.signedApiUrls] : acc), [] as string[])
+      );
+      // Overwrite the state with the new signed API URLs instead of merging them to avoid stale URLs.
+      updateState((draft) => {
```
bdrhn9: I wanted to re-write `signedApiUrls` of `state` entirely instead of updating only changing parts. The rationale is when we phase out an airnode, old implementation doesn't remove this from the state and will continue to fetch and save signed data from signed API. I thought that most effective way to achieve this purpose is updating state after all batches processed. That's why I moved `updateState` statement from the func `processBatch` to here.

src/update-feeds-loops/update-feeds-loops.ts
```diff
@@ -262,24 +271,26 @@ export const processBatch = async (
...
-      for (const { airnodeAddress, url } of receivedUrls) {
-        draft.signedApiUrls[chainId]![providerName]![airnodeAddress] = url;
-      }
-    }
-  });
+  // Generate signed API URLs for the batch
+  const signedApiUrls = batch
+    .map((dataFeed) =>
+      dataFeed.beaconsWithData.map((beacon, index) => {
+        const configSignedApiUrls = configSignedApiBaseUrls.map((baseUrl) => `${baseUrl}/${beacon.airnodeAddress}`);
```
bdrhn9, line 286: I preferred to generate all airnode-populated signedApiUrls here. So base signedApiUrl coming from config also utilized here. With doing this, I simplied the `data-fetcher-loop`. Now, its only responsibility is getting airnode-populated signed API URLs from the state and making the requests to them.

src/update-feeds-loops/update-feeds-loops.ts
Siegrift: > when we phase out an airnode

Not sure what this means.

src/state/state.test.ts
```diff
@@ -27,14 +27,7 @@ const stateMock: State = {
...
   },
-  signedApiUrls: {
-    '31337': {
-      hardhat: {
-        '0xC04575A2773Da9Cd23853A69694e02111b2c4182':
-          'http://127.0.0.1:8090/0xbF3137b0a7574563a23a8fC8badC6537F98197CC',
-      },
-    },
-  },
+  signedApiUrls: ['http://127.0.0.1:8090/0xC04575A2773Da9Cd23853A69694e02111b2c4182'],
```
Siegrift: I only took a quick look today, but this doesn't feel right to me. The reason why I stored the `signedApiUrls` as record of chain and provider was to handle potential different return values of the providers. We agreed that we will call the signed APIs returned by all providers across all chains (but each provider can report only a single signed API).

I'll take a proper look tomorrow.

src/update-feeds-loops/update-feeds-loops.ts
bdrhn9: For example, when we want to remove one of the partner API provider (e.g. Nodary), we won't use its signed data so we don't need to make signed api call for its airnode address. If you see [here](https://github.com/api3dao/airseeker-v2/blob/a99a98acb3a8235edb15aadf8c1274e77281c1a1/src/update-feeds-loops/update-feeds-loops.ts#L279), previous implementation only creates or updates the existing records instead of re-creating draft.signedApiUrls entirely.

src/state/state.test.ts
bdrhn9: > handle potential different return values of the providers.

I protected the same behaviour. I preferred arrays because no where utilizes records grouped by keys `chain` and `provider`. `Object.values(...)` leveraged to get all signedApiUrls, see old implementation [here](https://github.com/api3dao/airseeker-v2/blob/a99a98acb3a8235edb15aadf8c1274e77281c1a1/src/data-fetcher-loop/data-fetcher-loop.ts#L79).

src/data-fetcher-loop/signed-data-state.ts
```diff
@@ -101,7 +101,7 @@ export const purgeOldSignedData = () => {
   const state = getState();
   const oldSignedData = Object.values(state.signedDatas).filter((signedData) => isSignedDataFresh(signedData));
   if (oldSignedData.length > 0) {
-    logger.info(`Purging some old signed data.`, { oldSignedData });
+    logger.debug(`Purging some old signed data.`, { oldSignedData });
```
Siegrift, line 104: 👍 fyi: https://github.com/api3dao/airseeker-v2/issues/189 but I'd expect that there more logs to be changed, so I'd leave the issue open.

src/state/state.test.ts
Siegrift: Mhh, I still believe this won't work. We call the `runUpdateFeeds` [here](https://github.com/api3dao/airseeker-v2/blob/bdrhn9/issue190/src/update-feeds-loops/update-feeds-loops.ts#L54) for all providers across all chains concurrently.

So the signed API URLs [here](https://github.com/api3dao/airseeker-v2/blob/bdrhn9/issue190/src/update-feeds-loops/update-feeds-loops.ts#L246) come only from the single provider, but it removes the Signed APIs reported by others.

bdrhn9: You are right I missed that point 😞

Siegrift: No worries. That's pretty non-apparent behavior. I think we should cover that by unit test. Something along these lines:
```
  it('persists signed API URLs reported by different providers/chains', async () => {
    const dataFeed = generateActiveDataFeedResponse();
    const beacons = contractsModule.decodeDataFeedDetails(dataFeed.dataFeedDetails)!;
    const decodedUpdateParameters = contractsModule.decodeUpdateParameters(dataFeed.updateParameters);
    const activeDataFeed = {
      ...omit(dataFeed, ['dataFeedDetails', 'beaconValues', 'beaconTimestamps']),
      decodedUpdateParameters,
      beaconsWithData: contractsModule.createBeaconsWithData(beacons, dataFeed.beaconValues, dataFeed.beaconTimestamps),
      decodedDapiName: utilsModule.decodeDapiName(dataFeed.dapiName),
      signedApiUrls: [],
    };
    const testConfig = generateTestConfig();
    jest.spyOn(stateModule, 'getState').mockReturnValue(
      allowPartial<stateModule.State>({
        config: { ...testConfig, signedApiUrls: ['http://config.url'] },
        signedApiUrls: ['http://url-from-chain-reported-by-different-provider'],
      })
    );
    jest.spyOn(logger, 'warn');
    jest.spyOn(logger, 'info');

    // Skip actions other than generating signed api urls.
    jest.spyOn(getUpdatableFeedsModule, 'getUpdatableFeeds').mockReturnValue([]);
    jest.spyOn(submitTransactionModule, 'getDerivedSponsorWallet').mockReturnValue(ethers.Wallet.createRandom());
    jest.spyOn(submitTransactionModule, 'hasSponsorPendingTransaction').mockReturnValue(false);

    const { signedApiUrls } = await updateFeedsLoopsModule.processBatch(
      [activeDataFeed],
      'default-provider',
      new ethers.JsonRpcProvider(),
      '31337',
      123
    );

    expect(signedApiUrls).toHaveLength(2);
    // TODO: This breaks
    expect(signedApiUrls).toStrictEqual([
      'http://config.url/0xc52EeA00154B4fF1EbbF8Ba39FDe37F1AC3B9Fd4',
      'http://url-from-chain-reported-by-different-provider',
    ]);
  });
```

src/update-feeds-loops/update-feeds-loops.ts
bbenligiray: Another example where this could be relevant is if the API provider rotates their mnemonic periodically for security purposes

src/update-feeds-loops/update-feeds-loops.ts
```diff
@@ -247,9 +247,7 @@ export const runUpdateFeeds = async (providerName: ProviderName, chain: Chain, c
         processedBatches.reduce((acc, batch) => (batch ? [...acc, ...batch.signedApiUrls] : acc), [] as string[])
       );
       // Overwrite the state with the new signed API URLs instead of merging them to avoid stale URLs.
-      updateState((draft) => {
-        draft.signedApiUrls = signedApiUrls;
-      });
+      updateState((draft) => set(draft, ['signedApiUrls', chainId, providerName], signedApiUrls));
```
bdrhn9, line 250: I defined `state.signedApiUrls` as an object indexed by `chain` and `provider`, respectively.  The reason is that we are updating state in `runUpdateFeeds` which is called for each chain-provider pair. This approach not only facilitates the removal of deprecated signed API URLs but also ensures we don't inadvertently alter URLs belonging to other chain-provider groups.

src/data-fetcher-loop/data-fetcher-loop.ts
```diff
@@ -70,21 +70,21 @@ export const runDataFetcher = async () => {
...
-        .flat(),
-      ...signedApiUrls,
-    ]);
+    // Better to log the non-decomposed object to see which URL comes from which chain-provider group.
+    logger.debug('Fetching data from signed APIs.', { signedApiUrls });
+
+    const urls = uniq(
+      Object.values(signedApiUrls)
+        .map((urlsPerProvider) => Object.values(urlsPerProvider))
+        .flat(2)
```
Siegrift, line 85: Do we need to pass `2` as argument?

src/update-feeds-loops/update-feeds-loops.test.ts
Siegrift, line 1: Can we also add a test like this? https://github.com/api3dao/airseeker-v2/pull/193#discussion_r1507608573

I think it's going to be helpful for the future.

src/update-feeds-loops/update-feeds-loops.ts
```diff
@@ -262,24 +269,32 @@ export const processBatch = async (
...
-      }
-    }
-  });
+  // Generate signed API URLs for the batch
+  const signedApiUrls = batch
+    .map((dataFeed) =>
+      dataFeed.beaconsWithData.map((beacon, index) => {
+        const configSignedApiUrls = configSignedApiBaseUrls.map((baseUrl) => `${baseUrl}/${beacon.airnodeAddress}`);
+
+        // WARNING: contractSignedApiBaseUrl is an array of empty strings if it's not set on chain
```
Siegrift, line 288: I think you mean `signedApiUrls` and not `contractSignedApiBaseUrl`.

But I think it's better to reflect this information early on. It would be better to change the empty strings to `null` [here](https://github.com/api3dao/airseeker-v2/blob/21674d8a2e7cab44a39ff53ebb56830231863e7b/src/update-feeds-loops/contracts.ts#L153).

src/update-feeds-loops/update-feeds-loops.ts
Siegrift: I realised this has a small drawback that whenever processing the batch fails (e.g. temporary network error) we won't use the signed API URLs from the config. The previous approach had a different drawback - the Signed APIs for no longer active dAPIs were never cleaned up. 

I think the new behaviour is also OK, but I think it would deserve a comment.

Siegrift (approved):
Had some small suggestions, but otherwise 👍 LGTM

src/data-fetcher-loop/data-fetcher-loop.ts
bdrhn9: Yes, otherwise it is returning object in type string[][]. We are expecting `urls` object in string[].

src/update-feeds-loops/update-feeds-loops.test.ts
bdrhn9: Actually, I couldn't manage to add this test. Do we really need this test because we are now leveraging the object keyed by chain and provider pairs, it looks impossible to change non-related entry of the object, right?

Siegrift: Right. Makes sense.

PR #197, opened 2024-03-07
Fix release instructions

Siegrift:
Noticed small issues in the release instructions during 2.2.0 release.


Siegrift:
This is a simple README change, so I am merging right away.

PR #203, opened 2024-03-11
Set NODE_ENV to production, use node user

### LINKED ISSUE #199: Use "node" Docker user, use NODE_ENV=production
Similarly to https://github.com/api3dao/signed-api/pull/256

Siegrift:
Closes https://github.com/api3dao/airseeker-v2/issues/199

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -6,23 +6,21 @@ WORKDIR /app
 # Install pnpm globally (needs root privileges).
 RUN npm install -g pnpm
 
-# Create a new user and group to run the application with reduced privileges.
-RUN addgroup -S airseeker && \
-    adduser -h /app -s /bin/false -S -D -H -G airseeker airseeker && \
-    chown -R airseeker /app
-USER airseeker
+# "node" Docker images come with a built-in, least-privileged user called "node"
+USER node
 
 # Copy just the "pnpm-lock.yaml" file and use "pnpm fetch" to download all dependencies just from the lockfile. This
 # command is specifically designed to improve building a docker image because it only installs the dependencies if the
 # lockfile has changed (otherwise uses the cached value).
-COPY --chown=airseeker:airseeker pnpm-lock.yaml /app
+COPY --chown=node:node pnpm-lock.yaml /app
 RUN pnpm fetch
 # Copies all of the contents (without files listed in .dockerignore) of the monorepo into the image.
-COPY --chown=airseeker:airseeker . /app
+COPY --chown=node:node . /app
 # Ideally, we would use "--offline" option, but it seems pnpm has a bug. Fortunately, the instalation times are similar.
 # See: https://github.com/pnpm/pnpm/issues/6058 for details.
 RUN pnpm install --prefer-offline
 RUN pnpm run build
 
 # Start the Airseeker.
+ENV NODE_ENV=production
 ENTRYPOINT ["node", "dist/index.js"]
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -18,11 +18,9 @@ COPY . /app
 RUN pnpm install --prefer-offline
 RUN pnpm run build
 
-RUN chown -R node:node /app/config
-
+RUN chown --recursive node:node /app
 # "node" Docker images come with a built-in, least-privileged user called "node"
 USER node
-
 # Start the Airseeker.
 ENV NODE_ENV=production
 ENTRYPOINT ["node", "dist/index.js"]
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -6,20 +6,22 @@ WORKDIR /app
 # Install pnpm globally (needs root privileges).
 RUN npm install -g pnpm
 
-# "node" Docker images come with a built-in, least-privileged user called "node"
-USER node
-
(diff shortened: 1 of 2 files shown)


docker/Dockerfile
```diff
@@ -6,23 +6,21 @@ WORKDIR /app
...
-USER airseeker
+# "node" Docker images come with a built-in, least-privileged user called "node"
+USER node
 
 # Copy just the "pnpm-lock.yaml" file and use "pnpm fetch" to download all dependencies just from the lockfile. This
 # command is specifically designed to improve building a docker image because it only installs the dependencies if the
 # lockfile has changed (otherwise uses the cached value).
-COPY --chown=airseeker:airseeker pnpm-lock.yaml /app
+COPY --chown=node:node pnpm-lock.yaml /app
 RUN pnpm fetch
```
bdrhn9, line 13: When I try to build the image I got the following error at `RUN pnpm fetch` step:

```
 => [3/8] RUN npm install -g pnpm                                                                                                         1.7s
 => [4/8] COPY --chown=node:node pnpm-lock.yaml /app                                                                                      0.0s
 => ERROR [5/8] RUN pnpm fetch                                                                                                            0.9s
------                                                                                                                                         
 > [5/8] RUN pnpm fetch:
0.898  EACCES  EACCES: permission denied, open '/app/_tmp_1_5a8578c4de70d1f92035f168d4c8f1cf'
0.898 
0.898 
0.898 
------
Dockerfile:16
--------------------
  14 |     # lockfile has changed (otherwise uses the cached value).
  15 |     COPY --chown=node:node pnpm-lock.yaml /app
  16 | >>> RUN pnpm fetch
  17 |     # Copies all of the contents (without files listed in .dockerignore) of the monorepo into the image.
  18 |     COPY --chown=node:node . /app
--------------------
ERROR: failed to solve: process "/bin/sh -c pnpm fetch" did not complete successfully: exit code: 1
```

Solution would be installing all required packages as a root in the Dockerfile, and at the end of the file switching to image user to `node` by just adding `USER node` like following:
```dockerfile
FROM node:18-alpine
ENV name="airseeker"
LABEL application="airseeker" description="Airseeker container"
WORKDIR /app

# Install pnpm globally (needs root privileges).
RUN npm install -g pnpm

# Copy just the "pnpm-lock.yaml" file and use "pnpm fetch" to download all dependencies just from the lockfile. This
# command is specifically designed to improve building a docker image because it only installs the dependencies if the
# lockfile has changed (otherwise uses the cached value).
COPY pnpm-lock.yaml /app
RUN pnpm fetch
# Copies all of the contents (without files listed in .dockerignore) of the monorepo into the image.
COPY . /app
# Ideally, we would use "--offline" option, but it seems pnpm has a bug. Fortunately, the instalation times are similar.
# See: https://github.com/pnpm/pnpm/issues/6058 for details.
RUN pnpm install --prefer-offline
RUN pnpm run build

# "node" Docker images come with a built-in, least-privileged user called "node"
USER node

# Start the Airseeker.
ENV NODE_ENV=production
ENTRYPOINT ["node", "dist/index.js"]
```

That would work (even though `node` user isn't owner of the `/app` directory, `others` have right to read files under `/app`) if we don't utilize some trick to not read config file from S3 on Cloudformation like this:
```
mkdir -p /app/config && echo -e $AIRSEEKER_JSON > /app/config/airseeker.json && IFS= && echo -e $SECRETS_ENV > /app/config/secrets.env && node /app/dist/index.js
```

My solution would be adding this line before setting user `RUN chown -R node:node /app/config`. We can also choose `RUN chown -R node:node /app`. If we choose first, we can guarantee that we sealed all packages except config folder which is one we can expect to be changed dynamically. So:

```dockerfile
FROM node:18-alpine
ENV name="airseeker"
LABEL application="airseeker" description="Airseeker container"
WORKDIR /app

# Install pnpm globally (needs root privileges).
RUN npm install -g pnpm

# Copy just the "pnpm-lock.yaml" file and use "pnpm fetch" to download all dependencies just from the lockfile. This
# command is specifically designed to improve building a docker image because it only installs the dependencies if the
# lockfile has changed (otherwise uses the cached value).
COPY pnpm-lock.yaml /app
RUN pnpm fetch
# Copies all of the contents (without files listed in .dockerignore) of the monorepo into the image.
COPY . /app
# Ideally, we would use "--offline" option, but it seems pnpm has a bug. Fortunately, the instalation times are similar.
# See: https://github.com/pnpm/pnpm/issues/6058 for details.
RUN pnpm install --prefer-offline
RUN pnpm run build

RUN chown -R node:node /app/config

# "node" Docker images come with a built-in, least-privileged user called "node"
USER node

# Start the Airseeker.
ENV NODE_ENV=production
ENTRYPOINT ["node", "dist/index.js"]
```

bdrhn9 (changes requested):

docker/Dockerfile
Siegrift: Thanks. I actually haven't built it myself, because I thought it's going to be check by CI, but we only have that in Signed API repo 🤦 .

bdrhn9 (approved):

docker/Dockerfile
Siegrift: I've made a slight change to chown over the whole `/app` folder as it was done by this image previosuly (`chown -R airseeker /app`). This is the same way how it's done in Signed API.

PR #204, opened 2024-03-12
Update engines

Siegrift:
The engines are too strict and they are annoying when building new Docker image (because Docker caches the pnpm version) and this can also issue a warning when the package is installed as NPM.


bdrhn9 (approved):
Ah yes, I also failed while building docker image. I solved it by adding `--no-cache` option to docker build command. I'd add this option to script in package.json 😄

Siegrift:
>  I solved it by adding --no-cache option to docker build command. I'd add this option to script in package.json 😄

Yeah, I also used that a few times, but didn't want to add it to the script, but it was't a strong opinion. It won't matter much now, but I'll add it.

PR #211, opened 2024-03-20
Fix ethers provider draining resources in case of failure

### LINKED ISSUE #210: Investigate faulty RPC crashing Airseeker
See: https://api3workspace.slack.com/archives/C05RMQJ3X2L/p1710849365824949?thread_ts=1710845451.000169&cid=C05RMQJ3X2L

There is some unhandled error caused by faulty RPC.

Siegrift:
Closes https://github.com/api3dao/airseeker-v2/issues/210

## Rationale

I made a function which creates the provider in safe way and tested it locally. I also [let ethers know of this incident](https://github.com/api3dao/airseeker-v2/issues/210#issuecomment-2009989322) as I think the implementation could change a little.

### DIFF
--- a/src/update-feeds-loops/contracts.ts
+++ b/src/update-feeds-loops/contracts.ts
@@ -3,11 +3,35 @@ import {
   AirseekerRegistry__factory as AirseekerRegistryFactory,
   Api3ServerV1__factory as Api3ServerV1Factory,
 } from '@api3/contracts';
+import { go } from '@api3/promise-utils';
 import { ethers } from 'ethers';
 import { zip } from 'lodash';
 
 import type { AirnodeAddress, TemplateId } from '../types';
 import { decodeDapiName, deriveBeaconId } from '../utils';
+
+export const createProvider = async (url: string) => {
+  // Create a static network provider (bound to a specific network) to avoid the overhead of fetching the chain ID
+  // periodically.
+  const provider = new ethers.JsonRpcProvider(url, undefined, {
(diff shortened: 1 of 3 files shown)


bdrhn9 (approved):
Thanks for addressing this issue, it looks good 🔥

src/update-feeds-loops/contracts.ts
```diff
@@ -3,12 +3,36 @@ import {
...
   Api3ServerV1__factory as Api3ServerV1Factory,
 } from '@api3/contracts';
+import { go } from '@api3/promise-utils';
 import { ethers } from 'ethers';
 import { zip } from 'lodash';
 
 import type { AirnodeAddress, TemplateId } from '../types';
 import { decodeDapiName, deriveBeaconId } from '../utils';
 
+export const createProvider = async (url: string) => {
```
Siegrift, line 13: @aquarat provided a nicer alternative to this with the following gist:
```
const fetchClient = new FetchRequest(chain.rpcUrl);
  fetchClient.timeout = 10_000;

  const network = {
    name: chain.name,
    chainId: chain.chainId,
  };
  const providerPublic = new JsonRpcProvider(fetchClient, network, 
  {
    staticNetwork: Network.from(network.chainId), // or just network
  });
```

We know the chainID, but not the chain name. We could use `@api3/chains` for this, but I quite like that we "detect" the faulty providers this way, so I am keeping it this way. Shout if you have some strong preferences.

PR #214, opened 2024-03-21
Revisit INFO logs

### LINKED ISSUE #189: Revisit INFO logs
There are some logs which are too verbose ([example](https://github.com/api3dao/airseeker-v2/blob/main/src/data-fetcher-loop/signed-data-state.ts#L104)). They should be removed, changed severity or stripped payload (whatever makes more sense).

### LINKED ISSUE #198: Log with chain alias instead of chain ID
For instance, [here](https://github.com/api3dao/airseeker-v2/blob/bfd8adebb9e6ddc7a4803d9809012cdc4536f213/src/update-feeds-loops/update-feeds-loops.ts#L143), we're logging using the chainId, which may not be very human-readable or user-friendly. It would be more beneficial to utilize chain aliases for logging purposes instead.

Siegrift:
Closes https://github.com/api3dao/airseeker-v2/issues/189
Closes https://github.com/api3dao/airseeker-v2/issues/198

## Rationale

The INFO logs logged very little. This is now changed because both "data fetching loop" and "update feeds loops" now log a message. I went through each log one-by-one and I updated the severity whenever I felt it's a good idea.

I also changed that we log the chain name instead of chain ID.

### DIFF
--- a/src/data-fetcher-loop/signed-data-state.ts
+++ b/src/data-fetcher-loop/signed-data-state.ts
@@ -59,7 +59,7 @@ export const verifySignedDataIntegrity = (signedData: SignedData) => {
 };
 
 export const saveSignedData = (signedData: SignedData) => {
-  const { airnode, templateId, signature, timestamp, encodedValue } = signedData;
+  const { airnode, templateId, timestamp } = signedData;
 
   if (!verifySignedDataIntegrity(signedData)) {
     return;
@@ -76,17 +76,11 @@ export const saveSignedData = (signedData: SignedData) => {
   }
 
   if (!isSignedDataFresh(signedData)) {
-    logger.debug('Skipping state update. The signed data value is older than 24 hours.');
+    // This should not happen under normal circumstances. Something must be off with the Signed API.
+    logger.warn('Skipping state update. The signed data value is older than 24 hours.');
     return;
   }
 
-  logger.debug(`Storing signed data.`, {
-    airnode,
-    templateId,
-    timestamp,
-    signature,
-    encodedValue,
-  });
   updateState((draft) => {
     draft.signedDatas[dataFeedId] = signedData;
   });
--- a/src/update-feeds-loops/update-feeds-loops.ts
+++ b/src/update-feeds-loops/update-feeds-loops.ts
@@ -151,7 +151,7 @@ export const runUpdateFeeds = async (providerName: ProviderName, chain: Chain, c
       // Create a provider and connect it to the AirseekerRegistry contract.
       const provider = await createProvider(providers[providerName]!.url);
       if (!provider) {
-        logger.info('Failed to create provider. This is likely an RPC issue.');
+        logger.warn('Failed to create provider. This is likely an RPC issue.');
         return;
       }
       const airseekerRegistry = getAirseekerRegistry(contracts.AirseekerRegistry, provider);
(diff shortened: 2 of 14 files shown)


src/data-fetcher-loop/signed-data-state.ts
```diff
@@ -76,17 +76,11 @@ export const saveSignedData = (signedData: SignedData) => {
   }
 
   if (!isSignedDataFresh(signedData)) {
-    logger.debug('Skipping state update. The signed data value is older than 24 hours.');
+    // This should not happen under normal circumstances. Something must be off with the Signed API.
+    logger.warn('Skipping state update. The signed data value is older than 24 hours.');
     return;
   }
 
-  logger.debug(`Storing signed data.`, {
```
Siegrift, line 83: This is extremely verbose, so I'd just remove that.

src/update-feeds-loops/update-feeds-loops.ts
```diff
@@ -239,7 +239,7 @@ export const runUpdateFeeds = async (providerName: ProviderName, chain: Chain, c
       const skippedBatchesCount = processedBatches.filter((batch) => !batch).length;
       const dataFeedUpdates = processedBatches.reduce((acc, batch) => acc + (batch ? batch.successCount : 0), 0);
       const dataFeedUpdateFailures = processedBatches.reduce((acc, batch) => acc + (batch ? batch.errorCount : 0), 0);
-      logger.debug(`Finished processing batches of active data feeds.`, {
+      logger.info(`Finished processing batches of active data feeds.`, {
```
Siegrift: This will be logged at the end of the "update feeds loop" cycle. There will be no message at the start of i, but I think thats fine.

src/update-feeds-loops/update-feeds-loops.ts
```diff
@@ -171,7 +171,7 @@ export const runUpdateFeeds = async (providerName: ProviderName, chain: Chain, c
         if (goFirstBatch.data === null) return;
         const { batch: firstBatch, activeDataFeedCount, blockNumber: firstBatchBlockNumber } = goFirstBatch.data;
         if (activeDataFeedCount === 0) {
-          logger.warn(`No active data feeds found.`);
+          logger.info(`No active data feeds found.`);
```
Siegrift, line 174: This is too noisy in Grafana and is an expected (or better, possible) scenario so I downgraded this to INFO. I still think it's a valuable information so I don't want to decrease to DEBUG.

src/utils.ts
```diff
@@ -84,3 +85,7 @@ export const decodeBeaconValue = (encodedBeaconValue: string) => {
 
   return decodedBeaconValue;
 };
+
+export const getChainName = (chainId: string) => {
+  return CHAINS.find((c) => c.id === chainId)?.name ?? 'unknown';
```
bdrhn9:
```suggestion
  return CHAINS.find((c) => c.id === chainId)?.alias ?? 'unknown';
```
I'd rather to utilize chainAlias instead of long chain name in the logs.

bdrhn9 (approved):
There are some suggestions below. Other than that LGTM 🔥 

- [Here](https://github.com/api3dao/airseeker-v2/blob/better-logs/src/deviation-check/deviation-check.ts#L68) we can set log level to INFO, because it is similar to `Deviation exceed.` log above. 
- [Here](https://github.com/api3dao/airseeker-v2/blob/better-logs/src/gas-price/gas-price.ts#L189) can be set to WARN.

Siegrift:
> [Here](https://github.com/api3dao/airseeker-v2/blob/better-logs/src/gas-price/gas-price.ts#L189) can be set to WARN.

FYI, a bit related https://github.com/api3dao/airseeker-v2/issues/218

PR #216, opened 2024-03-21
Add stage and version

### LINKED ISSUE #212: Add version and stage
We should be able to identify Airseeker by custom ID (`stage`) and know the currently deployed version (`version`).

Siegrift:
Closes https://github.com/api3dao/airseeker-v2/issues/212

## Rationale

Inspired by https://github.com/api3dao/signed-api/blob/minor-cleanup/README.md#versioning-and-release and related code.


config/airseeker.example.json
```diff
@@ -25,5 +25,7 @@
   "deviationThresholdCoefficient": 1,
   "signedDataFetchInterval": 10,
   "signedApiUrls": [],
-  "walletDerivationScheme": { "type": "managed" }
+  "walletDerivationScheme": { "type": "managed" },
+  "stage": "dev",
+  "version": "2.2.1"
```
bdrhn9, line 30: Can we add `version` and `stage` fields to heartbeat payload? If you don't mind, it would be better to have `deploymentTimestamp` either 😄 . Not urgent.

README.md
```diff
@@ -247,6 +242,16 @@ The following options are available:
 - `managed` - Derives the wallet from the hash of the dAPI name (or data feed ID). This means the wallet derivation is
   agnostic to update parameters, and the same wallet is used when the dAPI is upgraded/downgraded.
 
+#### `stage`
+
+An identifier of the deployment stage. This is used to distinguish between different deployments, for example `dev`,
+`prod-1` or `prod-2`. The stage value can have 256 characters at maximum and can only include lowercase alphanumeric
+characters and hyphens.
+
+#### `version`
```
bdrhn9, line 251: Just a heads-up, we are using `version` on `signed-api`, `nodeVersion` on `airnode-feed`. At some point, it'd be good to be consistent on this.

bdrhn9 (approved):
LGTM, thanks 👍🏼

README.md
Siegrift: We might, but I don't think it's that big of a deal. Personally I prefer using `version`, but `nodeVersion` comes from Airnode, where it was named to distinguish it with `ois.version` and `ois.oisFormat`. I think for other services where there is no ambiguity it's OK to just use `version`.

config/airseeker.example.json
Siegrift: Sure, I was thinking that they are a bit unnecessary because there is the config hash. But yeah, you won't know what's the stage and version in Grafana logs without logging them in raw form. I'll change this.

PR #219, opened 2024-03-27
Avoid ethers code for possible infinite while loop

### LINKED ISSUE #210: Investigate faulty RPC crashing Airseeker

Siegrift:
Closes https://github.com/api3dao/airseeker-v2/issues/210

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,6 +148,11 @@ A record of chain configurations. The record key is the chain ID. For example:
 }
 ```
 
+##### `alias` _(optional)_
+
+The optional alias to identify the chain in logs. If not provided, the chain alias from `@api3/chains` is used. If there
+is no record in `@api3/chains` for the given chain ID, falls back to using `unknown`.
+
 ##### `contracts`
 
 A record of contract addresses used by Airseeker.
--- a/src/update-feeds-loops/contracts.ts
+++ b/src/update-feeds-loops/contracts.ts
@@ -10,25 +10,21 @@ import { zip } from 'lodash';
 import type { AirnodeAddress, TemplateId } from '../types';
 import { decodeDapiName, deriveBeaconId } from '../utils';
 
-export const createProvider = async (url: string) => {
+export const createProvider = async (chainId: string, chainAlias: string, providerUrl: string) => {
   // Create a static network provider (bound to a specific network) to avoid the overhead of fetching the chain ID
   // periodically.
-  const provider = new ethers.JsonRpcProvider(url, undefined, {
-    staticNetwork: true,
+  //
+  // Specifying the chain statically also avoids the code path that can cause infinite retries when the RPC is down.
+  // This is extremely unwanted behaviour, because Airseeker initializes provider in every cycle. See:
+  // https://github.com/ethers-io/ethers.js/blob/ad5f1c5fc7b73cfb20d9012b669e9dcb9122d244/src.ts/providers/provider-jsonrpc.ts#L762
+  const network = new ethers.Network(chainAlias, chainId);
+  const provider = new ethers.JsonRpcProvider(providerUrl, network, {
+    staticNetwork: network,
   });
 
-  // Unfortunately, ethers internally detects the network and when it fails it retries indefinitely (and logs and
-  // annoying message). See:
-  // https://github.com/ethers-io/ethers.js/blob/ad5f1c5fc7b73cfb20d9012b669e9dcb9122d244/src.ts/providers/provider-jsonrpc.ts#L762
-  //
-  // This is extremely unwanted behaviour, because if there is a broken RPC this will drain resources. Moreover, in
-  // Airseeker we initialize provider in every cycle. The workaround is to detect the network manually to ensure the RPC
-  // is healthy. If this is not the case, we destroy the provider stopping it from draining any resources.
-  const goDetectNetwork = await go(async () => provider._detectNetwork());
-  if (!goDetectNetwork.success) {
-    provider.destroy();
-    return null;
-  }
+  // Make sure the RPC is working by making a call to get the chain ID.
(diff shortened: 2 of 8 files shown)


README.md
```diff
@@ -148,6 +148,11 @@ A record of chain configurations. The record key is the chain ID. For example:
 }
 ```
 
+##### `alias` _(optional)_
```
Siegrift, line 151: I need the chain name in order to construct the `ethers.Network` object so I figured it might be nice to have this in the config. It uses the alias from `@api3/chains`, but can be used as override (and for HH during dev/testing).

src/update-feeds-loops/contracts.ts
```diff
@@ -10,25 +10,23 @@ import { zip } from 'lodash';
...
-  // Airseeker we initialize provider in every cycle. The workaround is to detect the network manually to ensure the RPC
-  // is healthy. If this is not the case, we destroy the provider stopping it from draining any resources.
-  const goDetectNetwork = await go(async () => provider._detectNetwork());
-  if (!goDetectNetwork.success) {
-    provider.destroy();
-    return null;
-  }
+  // This is extremely unwanted behaviour, because Airseeker initializes provider in every cycle.
+  const network = new ethers.Network(chainAlias, chainId);
+  const provider = new ethers.JsonRpcProvider(providerUrl, network, {
```
Siegrift, line 21: This time, I made sure the problematic code path is not called at all. This is because the provider is initialized with a `Network` object, which makes [this condition](https://github.com/ethers-io/ethers.js/blob/ad5f1c5fc7b73cfb20d9012b669e9dcb9122d244/src.ts/providers/provider-jsonrpc.ts#L762) return false and the while loop never runs.

src/update-feeds-loops/contracts.ts
```diff
@@ -10,25 +10,23 @@ import { zip } from 'lodash';
...
+    staticNetwork: network,
+  });
+
+  // Make sure the RPC is working by making a call to get the chain ID.
+  const goChainId = await go(async () => provider.send('eth_chainId', []));
```
Siegrift, line 26: I still prefer to make the chain ID call to detect whether the RPC is faulty, but this is very subjective and I am open to removing this in case anyone objects.

src/update-feeds-loops/update-feeds-loops.ts
```diff
@@ -141,16 +141,16 @@ export const readActiveDataFeedBatch = async (
 
 export const runUpdateFeeds = async (providerName: ProviderName, chain: Chain, chainId: ChainId) => {
   await logger.runWithContext(
-    { chainName: getChainName(chainId), providerName, updateFeedsCoordinatorId: Date.now().toString() },
+    { chainName: chain.alias, providerName, updateFeedsCoordinatorId: Date.now().toString() },
```
Siegrift, line 144: I could rename this from `chainName` to `alias`. WDYT?

src/update-feeds-loops/contracts.ts
bdrhn9: It will be redundant because [this](https://github.com/api3dao/airseeker-v2/blob/72af392da2e22ef192ef33004b124aafe87cb511/src/update-feeds-loops/update-feeds-loops.ts#L111).

However, it's just one time call at the beginning. I would also prefer to keep it.

bdrhn9 (approved):
To be honest, I didn't go deep by investigating about how Network and JsonRpcProvider classes implemented in ethers-v6 😄. However, changes looks good to me.

Siegrift:
> It will be redundant because [this](https://github.com/api3dao/airseeker-v2/blob/72af392da2e22ef192ef33004b124aafe87cb511/src/update-feeds-loops/update-feeds-loops.ts#L111).

That ensures that we've read the contract on the correct chain while the RPC call when creating provider ensures that the provider is working, so they serve different purposes.

yarikMyroniuk:
> provider.send('eth_chainId', [])

**Hello**

I was checking how you resolve this issue but when I changed my code (example below)

```    
     let provider
      try {
          const network = new ethers.Network(network.name, network.chainId);
          provider = new ethers.JsonRpcProvider(rpcUrlData.url, network, {
            staticNetwork: network,
          });
          
          await provider.send('eth_chainId', []);
      } catch (e) {
          if (provider) {
            provider.destroy();
          }
          throw e;
      } 
```

  I still have ```JsonRpcProvider failed to detect network and cannot start up; retry in 1s (perhaps the URL is wrong or the node is not started)```
  
  Can you help me please?
  @Siegrift 
  @bdrhn9

Siegrift:
Hi, that is the code we're still using and works for us. Not sure about the code snippet you've shared, but the `provider` inside the `catch` statement does not exist in scope - maybe that's the cause of your issue?

bdrhn9:
Your code is explicitly calling `await provider._detectNetwork();`. Please try with removing that as our code doesn't include it.

yarikMyroniuk:
Thank you so much guys for your quickly reply

I adjusted my code above, please check it

But I still have this log

JsonRpcProvider failed to detect network and cannot start up; retry in 1s (perhaps the URL is wrong or the node is not started)

UPD: 
   my ethers version: "ethers": "^6.9.1",

@bdrhn9 
@Siegrift

yarikMyroniuk:
Hello

Can you check above my code please?

Because I still have this log

@bdrhn9 
@Siegrift

bbenligiray:
@yarikMyroniuk feel free to create a standalone issue that describes your problem. It's not clear what you're after here.

PR #224, opened 2024-04-01
Fix gas price scaling

### LINKED ISSUE #223: Fix gas price scaling
[This check](https://github.com/api3dao/airseeker-v2/blob/e335a8579e851d5ea6db11d363eb329ac15ebc98/src/gas-price/gas-price.ts#L164) is wrong and prevents gas scaling.

Revisit the spec as well, because this is not as clear as it should. Also, the documentation for config fields provides very little detail.

Siegrift:
Closes https://github.com/api3dao/airseeker-v2/issues/223

## Rationale

There was a discrepancy in the implementation, originating from misalignment on how the gas oracle should be implemented.

### DIFF
--- a/src/gas-price/gas-price.ts
+++ b/src/gas-price/gas-price.ts
@@ -141,7 +141,7 @@ export const getRecommendedGasPrice = async (
   if (!goGasPrice.success) logger.error('Failed to fetch gas price from RPC provider.', goGasPrice.error);
   if (gasPrice) saveGasPrice(chainId, providerName, gasPrice);
 
-  // If the gas price from RPC provider is not available, use the last saved gas price (provided it's fresh enough)
+  // If the gas price from RPC provider is not available, use the last saved gas price.
   if (!gasPrice && gasPrices.length > 0) {
     const lastSavedTimestamp = Math.max(...gasPrices.map((gasPrice) => gasPrice.timestamp));
     const lastSavedGasPrice = gasPrices.find((gasPrice) => gasPrice.timestamp === lastSavedTimestamp)!.price;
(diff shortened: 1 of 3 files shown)


src/gas-price/gas-price.ts
```diff
@@ -161,15 +161,16 @@ export const getRecommendedGasPrice = async (
...
     const multiplier = calculateScalingMultiplier(
       recommendedGasPriceMultiplier,
       maxScalingMultiplier,
-      Math.floor(Date.now() / 1000) - lastUpdateTimestamp,
+      pendingPeriod,
       scalingWindow
     );
 
-    logger.warn('Scaling gas price.', { gasPrice: gasPrice.toString(), multiplier });
+    logger.warn('Scaling gas price.', { gasPrice: gasPrice.toString(), multiplier, pendingPeriod });
```
Siegrift, line 173: I've added a `pendingPeriod` field, because it's quite important to know for how long the transaction is pending.

src/gas-price/gas-price.ts
```diff
@@ -161,15 +161,16 @@ export const getRecommendedGasPrice = async (
   const lastUpdateTimestamp = sponsorLastUpdateTimestamp[sponsorWalletAddress];
 
   // Check if the next update is a retry of a pending transaction and if it has been pending longer than scalingWindow
```
vponline:
```suggestion
  // Check if the next update is a retry of a pending transaction
```

vponline (approved):
LGTM, Thanks for fixing!

src/gas-price/gas-price.ts
```diff
@@ -161,15 +161,16 @@ export const getRecommendedGasPrice = async (
...
-  if (lastUpdateTimestamp && lastUpdateTimestamp < Math.floor(Date.now() / 1000) - scalingWindow) {
+  if (lastUpdateTimestamp) {
+    const pendingPeriod = Math.floor(Date.now() / 1000) - lastUpdateTimestamp;
     const multiplier = calculateScalingMultiplier(
       recommendedGasPriceMultiplier,
       maxScalingMultiplier,
```
bdrhn9, line 168: It's nit and not related to objective of this PR but I'd suggest renaming `maxScalingMultiplier` to `maxGasPriceMultiplier` or `maxRecommendedGasPriceMultiplier`. The current name, `maxScalingMultiplier`, doesn't convey the intended meaning effectively; it could be misinterpreted as a multiplier for `recommendedGasPriceMultiplier`.

bdrhn9 (approved):
It looks good to me 🔥

src/gas-price/gas-price.ts
Siegrift: Hehe, I actually was thinking similarly [here](https://github.com/api3dao/airseeker-v2/pull/224/commits/66cf9152da5989f87a86a1a3fa29e1bcf1f95d86#diff-a25ae601baae9bca444c00d66e4934596768a26eafd5208c8fa077dcf31b4ae9R40)
 but at the end didn't do it, because it would be a breaking change and I wasn't sure if I am not too subjective.
 
At the end I also accepted `maxScalingMultiplier` as a good name, because it's only applied during gas price scaling. I still think `baseGasPriceMultiplier` would be a better name compared to `recommendedGasPriceMultiplier`

Siegrift: I feel like it's not worth doing (nor opening an issue for). If you feel otherwise, feel free to create the issue.

src/gas-price/gas-price.ts
Siegrift: Thanks.

PR #225, opened 2024-04-01
Fix gas fetching, improve tests

### LINKED ISSUE #218: Revisit how gas prices for sanitization are fetched
We currently fetch the gas price only when submitting update tx. This is quite rare and we end up with not having enough gas price data to sanitize gas prices.

See: https://api3workspace.slack.com/archives/C06R2JCFSS1/p1711559821427089?thread_ts=1711546852.953019&cid=C06R2JCFSS1

Siegrift:
Closes https://github.com/api3dao/airseeker-v2/issues/218

## Rationale

[Here](https://api3workspace.slack.com/archives/C05TQPT7PNJ/p1697793195536069) we've decided to make the gas fetching part of the "main" (update feeds) loop. The problem is that the gas was fetched from the RPC only before an update transaction was submitted.

Now it's fetched during the processing of each batch.

### DIFF
--- a/src/gas-price/gas-price.ts
+++ b/src/gas-price/gas-price.ts
@@ -158,7 +158,7 @@ export const getRecommendedGasPrice = (chainId: string, providerName: string, sp
   let latestGasPrice: bigint | undefined;
   // Use the latest gas price that is stored in the state. We assume that the gas price is fetched frequently and has
   // been fetched immediately before making this call. In case it fails, we fallback to the previously stored gas price.
-  if (!latestGasPrice && gasPrices.length > 0) {
+  if (gasPrices.length > 0) {
     const lastSavedTimestamp = Math.max(...gasPrices.map((gasPrice) => gasPrice.timestamp));
     latestGasPrice = gasPrices.find((gasPrice) => gasPrice.timestamp === lastSavedTimestamp)!.price;
   }
--- a/src/gas-price/gas-price.ts
+++ b/src/gas-price/gas-price.ts
@@ -1,6 +1,6 @@
 import { go } from '@api3/promise-utils';
 import type { ethers } from 'ethers';
-import { remove } from 'lodash';
+import { maxBy, remove } from 'lodash';
 
 import { logger } from '../logger';
 import { getState, updateState } from '../state';
@@ -117,7 +117,7 @@ export const fetchAndStoreGasPrice = async (
   });
   const gasPrice = goGasPrice.data;
   if (!goGasPrice.success) {
-    if (!goGasPrice.success) logger.error('Failed to fetch gas price from RPC provider.', goGasPrice.error);
+    logger.error('Failed to fetch gas price from RPC provider.', goGasPrice.error);
     return null;
   }
   if (!gasPrice) {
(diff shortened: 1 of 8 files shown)


src/gas-price/gas-price.ts
```diff
@@ -185,28 +193,23 @@ export const getRecommendedGasPrice = async (
   const percentileGasPrice = getPercentile(
     sanitizationPercentile,
     gasPrices.map((gasPrice) => gasPrice.price)
-  );
-  if (!percentileGasPrice) {
-    logger.warn('No historical gas prices to compute the percentile. Using the provider recommended gas price.');
-    return multiplyBigNumber(gasPrice, recommendedGasPriceMultiplier);
-  }
-
+  )!;
```
Siegrift, line 193: Guaranteed to exist because the array is non-empty.

src/gas-price/gas-price.ts
```diff
@@ -102,6 +102,38 @@ export const calculateScalingMultiplier = (
...
+    const feeData = await provider.getFeeData();
+    // We assume the legacy gas price will always exist. See:
+    // https://api3workspace.slack.com/archives/C05TQPT7PNJ/p1699098552350519
+    return feeData.gasPrice;
+  });
+  const gasPrice = goGasPrice.data;
+  if (!goGasPrice.success) {
+    if (!goGasPrice.success) logger.error('Failed to fetch gas price from RPC provider.', goGasPrice.error);
+    return null;
+  }
```
vponline: This seems like a typo:
```suggestion
  if (!goGasPrice.success) {
    logger.error('Failed to fetch gas price from RPC provider.', goGasPrice.error);
    return null;
  }
```

vponline (approved):
LGTM, noticed one possible typo

src/gas-price/gas-price.ts
```diff
@@ -126,40 +153,21 @@ export const getRecommendedGasPrice = async (
...
+  // Use the latest gas price that is stored in the state. We assume that the gas price is fetched frequently and has
+  // been fetched immediately before making this call. In case it fails, we fallback to the previously stored gas price.
+  if (gasPrices.length > 0) {
     const lastSavedTimestamp = Math.max(...gasPrices.map((gasPrice) => gasPrice.timestamp));
-    const lastSavedGasPrice = gasPrices.find((gasPrice) => gasPrice.timestamp === lastSavedTimestamp)!.price;
-
-    if (lastSavedTimestamp >= Math.floor(Date.now() / 1000) - 10 * dataFeedUpdateInterval) {
-      gasPrice = lastSavedGasPrice;
-    }
+    latestGasPrice = gasPrices.find((gasPrice) => gasPrice.timestamp === lastSavedTimestamp)!.price;
```
bdrhn9: Nit: I'd rather to utilize lodash's `maxBy` function like:
```typescript
latestGasPrice = _.maxBy(gasPrices, 'timestamp').price
```

src/update-feeds-loops/update-feeds-loops.ts
```diff
@@ -339,6 +339,10 @@ export const processBatch = async (
     }
   }
 
+  // Fetch the gas price regardless of whether there are any feeds to be updated or not in order for gas oracle to
+  // maintain historical gas prices.
+  await fetchAndStoreGasPrice(chainId, providerName, provider);
```
bdrhn9, line 344: Instead of here, isn't it better to call this function on `runUpdateFeeds` function? Because there is no meaningful time between processBatch calls and it will cause redundant RPC calls. For example, if we have 200 feeds on a chain and batch size 10, we probably requests same gas price 20 times for each RPC provider.

bdrhn9 (approved):
I liked the optimizations :-) thanks for addressing them 🔥

src/update-feeds-loops/update-feeds-loops.ts
Siegrift: The [batches are staggered](https://github.com/api3dao/airseeker-v2/blob/22a1ddae6585c2c8eb79f4f08ff36899e2cd2bb8/src/update-feeds-loops/update-feeds-loops.ts#L201) so there is a time delay. If we only fetched in in `runUpdateFeeds` the last batch would use gas price old ~`dataFeedUpdateInterval`.

 I see the point with multiple RPC calls though, but I don't see a better place when to fetch this.

bdrhn9: It makes sense. Your approach looks more proper 👍🏼

src/gas-price/gas-price.ts
Siegrift: Yeah, thanks.

src/gas-price/gas-price.ts
Siegrift: I dislike the shorthand property, because it has not a good TS support and can break when you rename the field. This is a bit more typesafe:
```
maxBy(gasPrices, (x) => x.timestamp)
```

PR #232, opened 2024-04-03
Fix dockerfile

### LINKED ISSUE #229: Update entrypoint in Dockerfile
As I figured out, adding the following line causes to compiling source files under `./dist/src/*` instead of `./dist/*`, that's why we need to update [entrypoint](https://github.com/api3dao/airseeker-v2/blob/aeea9d27eacdfe703d738efd194a1c8cb8daa94e/docker/Dockerfile#L26) of Dockerfile to be `/app/dist/src/index.js` 

https://github.com/api3dao/airseeker-v2/blob/2e8db23ab952e3cf21c19c307090c4ba529d00fe/tsconfig.json#L28

Siegrift:
Closes https://github.com/api3dao/airseeker-v2/issues/229

I tested this locally and it works. The change is similar to https://github.com/api3dao/signed-api/blob/7f30af2a5ada7abd7dc168da1032683b966410f8/packages/airnode-feed/src/validation/config.ts#L19

### DIFF
--- a/src/config/index.ts
+++ b/src/config/index.ts
@@ -1,5 +1,6 @@
 import { readFileSync } from 'node:fs';
 import { join } from 'node:path';
+import { cwd } from 'node:process';
 
 import { goSync } from '@api3/promise-utils';
 import dotenv from 'dotenv';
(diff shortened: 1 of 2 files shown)


src/config/index.ts
```diff
@@ -1,13 +1,18 @@
...
 import { goSync } from '@api3/promise-utils';
 import dotenv from 'dotenv';
 
 import { configSchema } from './schema';
 import { interpolateSecrets, parseSecrets } from './utils';
 
-export const getConfigPath = () => join(__dirname, '../../config');
+// When Airnode feed is built, the "/dist" file contains "src" folder and "package.json" and the config is expected to
+// be located next to the "/dist" folder. When run in development, the config is expected to be located next to the
+// "src" folder (one less import level). We resolve the config by CWD as a workaround. Since the Airnode feed is
```
bdrhn9:
```suggestion
// "src" folder (one less import level). We resolve the config by CWD as a workaround. Since the Airseeker is
```

Nit: typo.

bdrhn9 (approved):
Tested locally, it works as expected.

src/config/index.ts
Siegrift: I certainly did not copy paste the function from anywhere 👀

PR #233, opened 2024-04-04
Use deviation reference

### LINKED ISSUE #230: Respect the deviation reference
Update parameters includes a deviation reference. This should be respected while deciding if a deviation update needs to be made.

Siegrift:
Closes https://github.com/api3dao/airseeker-v2/issues/230

## Rationale

Mirror the contract implementation exactly. I believe this should exactly match the logic in the contract.


src/deviation-check/deviation-check.ts
```diff
@@ -1,14 +1,16 @@
...
-  const absoluteDelta = abs(delta);
-
-  // Avoid division by 0
-  const absoluteInitialValue = initialValue === 0n ? 1n : abs(initialValue);
+export const calculateDeviationPercentage = (
+  initialValue: bigint,
+  updatedValue: bigint,
+  deviationReference: bigint
+) => {
+  const absoluteDelta = abs(updatedValue - initialValue);
```
bdrhn9, line 10: After this line, we need to check that `absoluteDelta` is zero. If it is, we must return 0 deviation in percentage like in the [contract](https://github.com/api3dao/airnode-protocol-v1/blob/main/contracts/api3-server-v1/extensions/BeaconSetUpdatesWithPsp.sol#L162). One can say that if `absoluteDelta` is zero, it will return zero at the end of the function, however if  `absoluteInitialValue` is also zero, it will return infinite deviation in percentage.

src/deviation-check/deviation-check.ts
```diff
@@ -25,46 +27,67 @@ export const calculateMedian = (arr: bigint[]) => {
...
     if (offChainTimestamp < onChainTimestamp) {
       logger.warn(`Off-chain sample's timestamp is older than on-chain timestamp.`);
     }
     return false;
   }
 
-  // Check that on chain data is newer than heartbeat interval
+  // At this point it's clear that the off-chain data is newer than the on-chain data and the on-chain data is
+  // uninitialized.
+  if (onChainTimestamp === 0n) return true;
```
bdrhn9: To totally stick with [contract](https://github.com/api3dao/airnode-protocol-v1/blob/main/contracts/api3-server-v1/extensions/BeaconSetUpdatesWithPsp.sol#L126), maybe we can do:
```suggestion
  if (onChainTimestamp === 0n && offChainTimestamp !== 0n ) return true;
```

src/deviation-check/deviation-check.ts
```diff
@@ -25,46 +27,67 @@ export const calculateMedian = (arr: bigint[]) => {
...
+
+  // Check that on-chain data is fresh enough to not be force-updated by the heartbeat.
   const isFreshEnough = isOnChainDataFresh(onChainTimestamp, heartbeatInterval);
```
bdrhn9: We can do this check at the end as in the [contract](https://github.com/api3dao/airnode-protocol-v1/blob/main/contracts/api3-server-v1/extensions/BeaconSetUpdatesWithPsp.sol#L135). This will also make flow more readable.

src/deviation-check/deviation-check.ts
Siegrift: Right 🤦

src/deviation-check/deviation-check.ts
Siegrift: On a few lines above we check `if (offChainTimestamp <= onChainTimestamp)` so this is inferred.

Siegrift: But the check can't really happen, because we only update beacons that can be updated. So it doesn't make sense to check whether the off-chain value is smaller on-chain. The only reason that exists is because the test has inconsistent mock data.

I'll adhere to the contract style.

Siegrift (commented):
Thanks for catching that blunder, I need to be more thorough 😢 . 

At first I didn't want to make too many changes to the current style, but yeah rewriting it to contract style is much better 🙏 .

src/deviation-check/deviation-check.ts
Siegrift: Agree.

src/deviation-check/deviation-check.ts
```diff
@@ -25,46 +29,39 @@ export const calculateMedian = (arr: bigint[]) => {
...
   offChainTimestamp: bigint,
   heartbeatInterval: bigint,
-  deviationThreshold: bigint
+  deviationThreshold: bigint,
+  deviationReference: bigint
 ): boolean => {
-  // Check that fulfillment data is newer than on chain data. Update transaction with stale data would revert on chain,
-  // draining the sponsor wallet. See:
-  // https://github.com/api3dao/airnode-protocol-v1/blob/dev/contracts/dapis/DataFeedServer.sol#L121
-  if (offChainTimestamp <= onChainTimestamp) {
```
bdrhn9, line 51: We surely need this check. Because airseeker-v2 utilizes delayed endpoint, there can be some scenario like:
1. One updates data feed with OEV which is zero-delayed data
2. Airseeker has 15 secs delayed data which can exceed threshold
3. Airseeker tries to update feed but it reverts because [this](https://github.com/api3dao/contracts/blob/phase21/contracts/api3-server-v1/DataFeedServer.sol#L120) and it results in unnecessarily spent from sponsor wallet.

Siegrift: Before doing step 3) we check which beacons need to be updated [here](https://github.com/api3dao/airseeker-v2/blob/0ac8e98a4f90fb15dbc8b1f5fb20dabfc1ea1e85/src/update-feeds-loops/get-updatable-feeds.ts#L54) and  would use the on-chain value (provided we fetched it). In case we don't have the latest on-chain data, then Airseeker may attempt an update which would revert - this might fail [here](https://github.com/api3dao/airseeker-v2/blob/0ac8e98a4f90fb15dbc8b1f5fb20dabfc1ea1e85/src/update-feeds-loops/submit-transactions.ts#L168), but if there is a fallback gas limit provided, the transaction may indeed by submitted.

OEV updates are proxy specific, so that doesn't matter - but we may be running backup Airseekers so we should assume the feed can be updated from other sources, so your point is still valid.

But your comment made me think of another case, which I missed previously:
1. Say there is beacon set with 3 beacons.
2. Say Airseeker updates all beacons and the beacon set.
3. Now Signed API goes down, so there are no new data. This persists for a a couple of minutes.
4. None of the beacons are updatable, but the beacon set is because of the heartbeat.
5. We call [this](https://github.com/api3dao/airnode-protocol-v1/blob/65a77cdc23dc5434e143357a506327b9f0ccb7ef/contracts/api3-server-v1/DataFeedServer.sol#L46) without updating any beacon which will revert - although we could have prevent this.

Note that this is only for the beacon set, but yeah - we need the check for equality (but not for being smaller).

Siegrift: There is a similar edge case when the data feed is a beacon. What ends up happening we submit a multi call transaction with empty calldata.

Siegrift: Take a look at https://github.com/api3dao/airseeker-v2/pull/233/commits/6bd46b286769d89b595e7d22aa1c8344c04522a7

I think that mimics the contract logic. Note, that the timestamp check (as it was before) is more strict compared to the contract, because the contract allows the beacon set also if value change (and timestamp stays the same).

bdrhn9: I think that the most accurate implementation has been reached. I've thought about it quite and couldn't find anything to worry about 💪🏼.

src/update-feeds-loops/get-updatable-feeds.ts
```diff
@@ -54,15 +54,36 @@ export const getUpdatableFeeds = (
...
+        // 2. Data feed is a beacon set - We need make sure that the beacon set will change. The contract requires the
+        //    beacon set value or timestamp to change. See:
+        //    https://github.com/api3dao/airnode-protocol-v1/blob/65a77cdc23dc5434e143357a506327b9f0ccb7ef/contracts/api3-server-v1/DataFeedServer.sol#L54
+        const isBeaconSet = beaconValues.length > 1;
+        // Note, that the beacon set value/timestamp is computed as median, so single beacon update may not result in a beacon set update.
+        if (isBeaconSet && newDataFeedValue === dataFeedValue && newDataFeedTimestamp === dataFeedTimestamp) {
+          return false;
+        }
+        if (!isBeaconSet && newDataFeedTimestamp === dataFeedTimestamp) return false;
+
```
bdrhn9, line 90: I'm just nit-picking 😄. Because this snippet is directly related to `isDataFeedUpdatable`, it'd be better to place it into that function. Maybe that function can accept extra parameter `isBeaconSet`. 

Addition to that, for this condition `!isBeaconSet && newDataFeedTimestamp === dataFeedTimestamp`, it's hard for one to understand that this equals to `newDataFeedTimestamp <= dataFeedTimestamp` (coming from [here](https://github.com/api3dao/airseeker-v2/blob/6bd46b286769d89b595e7d22aa1c8344c04522a7/src/update-feeds-loops/get-updatable-feeds.ts#L41)). So maybe we can just write this condition explicitly again like `!isBeaconSet && newDataFeedTimestamp <= dataFeedTimestamp`.

bdrhn9 (approved):
Thank you for the PR! It's making significant improvements to update condition checks 🔥

src/update-feeds-loops/get-updatable-feeds.ts
Siegrift: > Because this snippet is directly related to isDataFeedUpdatable, it'd be better to place it into that function. Maybe that function can accept extra parameter isBeaconSet.

I was thinking about that and the problem I have with that was that it would stop mimicking the contract implementation of deviation logic. I think in the source of truth for the update condition is `getUpdatableFeeds` and not the deviation checks because even if we move the timestamp checks there, making sure we only update the beacons for which we have newer data is as important (and that happens outside of deviation checks).

> Addition to that, for this condition !isBeaconSet && newDataFeedTimestamp === dataFeedTimestamp, it's hard for one to understand that this equals to newDataFeedTimestamp <= dataFeedTimestamp (coming from [here](https://github.com/api3dao/airseeker-v2/blob/6bd46b286769d89b595e7d22aa1c8344c04522a7/src/update-feeds-loops/get-updatable-feeds.ts#L41)). So maybe we can just write this condition explicitly again like !isBeaconSet && newDataFeedTimestamp <= dataFeedTimestamp.

What I dislike about that is that it makes things more confusing once you understand what's happening - for example, the timestamp check I originally removed seemed extraneous because it was checking whether off-chain timestamp is smaller than on-chain which could never occur.

Siegrift: I tried to clarify this in the comments and updated the deviation checks function to stress it more.

PR #247, opened 2024-04-10
Wrap getting nonce in go

### LINKED ISSUE #242: Wrap getting nonce in go
https://github.com/api3dao/airseeker-v2/blob/3afaf311177d7c5e9271fefd948cf6631f789f78/src/update-feeds-loops/submit-transactions.ts#L85

We should wrap that in go utils. Related to that, we might want to revisit how update timestamp is persisted. See [here](https://api3workspace.slack.com/archives/C05TQPT7PNJ/p1712567895101259).

Siegrift:
Closes https://github.com/api3dao/airseeker-v2/issues/242

### DIFF
--- a/src/update-feeds-loops/submit-transactions.ts
+++ b/src/update-feeds-loops/submit-transactions.ts
@@ -82,7 +82,12 @@ export const submitTransaction = async (
         );
 
         logger.debug('Getting nonce.');
-        const nonce = await provider.getTransactionCount(sponsorWallet.address, blockNumber);
+        const goNonce = await go(async () => provider.getTransactionCount(sponsorWallet.address, 'latest'));
+        if (!goNonce.success) {
+          logger.warn(`Failed to get nonce.`, goNonce.error);
+          return null;
+        }
+        const nonce = goNonce.data;
 
         logger.debug('Getting recommended gas price.');
         const gasPrice = getRecommendedGasPrice(chainId, providerName, sponsorWallet.address);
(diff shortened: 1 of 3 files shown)


src/update-feeds-loops/submit-transactions.ts
```diff
@@ -82,7 +82,12 @@ export const submitTransaction = async (
         );
 
         logger.debug('Getting nonce.');
-        const nonce = await provider.getTransactionCount(sponsorWallet.address, blockNumber);
+        const goNonce = await go(async () => provider.getTransactionCount(sponsorWallet.address, 'latest'));
```
bdrhn9: We must use specific blockNumber to not cause any race-conditioning between different RPC providers or Airseeker-v2 instances. By doing [that](https://github.com/api3dao/airseeker-v2/blob/aae424814df3cd44d5aff99212ea1fa4fc0edd82/src/update-feeds-loops/update-feeds-loops.ts#L92), we are ensuring that whole loop is working for the same state on the chain.

bdrhn9 (approved):
Other than the comment which needs to be addressed, LGTM.

src/update-feeds-loops/submit-transactions.ts
Siegrift: 🤦 I have no idea how did I end up writing `latest` there.

Siegrift: I think it was because of my carelessness and Copilot...

PR #248, opened 2024-04-10
Try to detect submit tx failures

### LINKED ISSUE #243: Detect when update transaction fails with low balance
See: https://api3.grafana.net/goto/BFN9LW-Ig?orgId=1

We could log a nicer error message for this

Siegrift:
Closes https://github.com/api3dao/airseeker-v2/issues/243

## Rationale

There are other grafana [error codes](https://api3.grafana.net/explore?schemaVersion=1&panes=%7B%22ziv%22:%7B%22datasource%22:%22grafanacloud-logs%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22%7Bapp%3D%5C%22airseeker-v2%5C%22%7D%20%7C%3D%20%60Failed%20to%20submit%20the%20multicall%20transaction.%60%20%21%3D%20%60NONCE_EXPIRED%60%20%21%3D%20%60INSUFFICIENT_FUNDS%60%22,%22queryType%22:%22range%22,%22datasource%22:%7B%22type%22:%22loki%22,%22uid%22:%22grafanacloud-logs%22%7D,%22editorMode%22:%22builder%22%7D%5D,%22range%22:%7B%22from%22:%22now-30d%22,%22to%22:%22now%22%7D,%22panelsState%22:%7B%22logs%22:%7B%22logs%22:%7B%22visualisationType%22:%22logs%22%7D%7D%7D%7D%7D&orgId=1) but these are generic and rare. I don't want to do any message parsing so only relying on the error code.

I expect this will help with making the alert more actionable.

### DIFF
--- a/src/update-feeds-loops/submit-transactions.ts
+++ b/src/update-feeds-loops/submit-transactions.ts
@@ -118,7 +118,8 @@ export const submitTransaction = async (
         });
         if (!goMulticall.success) {
           // It seems that in practice, this code is widely used. We can do a best-effort attempt to determine the error
-          // reason. Many times, the error is acceptable and results from the way Airseeker is designed.
+          // reason. Many times, the error is acceptable and results from the way Airseeker is designed. We can use
+          // different log levels and messages and have better alerts.
           const errorCode = (goMulticall.error as any).code;
           switch (errorCode) {
             case 'REPLACEMENT_UNDERPRICED': {
--- a/src/update-feeds-loops/submit-transactions.ts
+++ b/src/update-feeds-loops/submit-transactions.ts
@@ -117,15 +117,28 @@ export const submitTransaction = async (
           );
         });
         if (!goMulticall.success) {
-          // It is possible (and sometimes expected) that we try to submit a replacement transaction with insufficient
-          // gas price. Because this is intended flow, we catch the transaction error and log an information message
-          // instead.
-          if ((goMulticall.error as any).code === 'REPLACEMENT_UNDERPRICED') {
-            logger.info(`Failed to submit replacement transaction because it was underpriced.`);
-            return null;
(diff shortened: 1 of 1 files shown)


src/update-feeds-loops/submit-transactions.ts
```diff
@@ -117,15 +117,28 @@ export const submitTransaction = async (
...
-          // It is possible (and sometimes expected) that we try to submit a replacement transaction with insufficient
-          // gas price. Because this is intended flow, we catch the transaction error and log an information message
-          // instead.
-          if ((goMulticall.error as any).code === 'REPLACEMENT_UNDERPRICED') {
-            logger.info(`Failed to submit replacement transaction because it was underpriced.`);
-            return null;
+          // It seems that in practice, this code is widely used. We can do a best-effort attempt to determine the error
+          // reason. Many times, the error is acceptable and results from the way Airseeker is designed.
+          const errorCode = (goMulticall.error as any).code;
+          switch (errorCode) {
```
bdrhn9, line 124: Probably I'd not try to make this error codes more verbose because I think they are self-explanatory. Furthermore, it's hard to cover all error codes, see [here](https://github.com/ethers-io/ethers.js/blob/78829050853093bc5291ae78fc5a904044759aa0/src.ts/utils/errors.ts#L68).

Siegrift: I don't want to cover all of them, but these are the most common ones we get. The main motivation is that we can have a better alert for this (or no alert - just an INFO message). 

Arguably we could do some ugly filtering in Grafana, but that would be ugly and quite hacky as well.

bdrhn9 (approved):
Implementation looks good to me.

PR #258, opened 2024-04-18
Gas sanitization

### LINKED ISSUE #245: Make gas sanitization median based
On today's Airseeker call, we talked about:

- Gas scaling using the latest gas price instead of sanitized one. When we sanitize and scale in short succession, we often end up overpaying.
- Current (percentile based) sanitization can suffer if last few minutes cause a gas spike.
- Sanitization could be based on the median of gas prices and multiplied by some factor and also cap the scaling.

The median could be parametrised, but in practice we will likely use only the median. There might be some edge cases, so this should be thought through a bit more. Even if the idea is sounds, it's best to see this work in practice.

Siegrift:
Closes https://github.com/api3dao/airseeker-v2/issues/245

### DIFF
--- a/local-test-configuration/airseeker/airseeker.example.json
+++ b/local-test-configuration/airseeker/airseeker.example.json
@@ -12,12 +12,12 @@
         }
       },
       "gasSettings": {
-        "recommendedGasPriceMultiplier": 1.5,
+        "recommendedGasPriceMultiplier": 1,
         "sanitizationSamplingWindow": 900,
         "sanitizationPercentile": 80,
         "scalingWindow": 120,
         "maxScalingMultiplier": 2,
-        "sanitizationMultiplier": 2
+        "sanitizationMultiplier": 3
       },
       "dataFeedUpdateInterval": 30,
       "dataFeedBatchSize": 10,
--- a/local-test-configuration/airseeker/airseeker.example.json
+++ b/local-test-configuration/airseeker/airseeker.example.json
@@ -16,7 +16,8 @@
         "sanitizationSamplingWindow": 900,
         "sanitizationPercentile": 80,
         "scalingWindow": 120,
-        "maxScalingMultiplier": 2
+        "maxScalingMultiplier": 2,
+        "sanitizationMultiplier": 2
       },
       "dataFeedUpdateInterval": 30,
       "dataFeedBatchSize": 10,
--- a/src/gas-price/gas-price.ts
+++ b/src/gas-price/gas-price.ts
@@ -107,7 +107,7 @@ export const getRecommendedGasPrice = (chainId: string, providerName: string, sp
   }
 
   // Check if the next update is a retry of a pending transaction and scale the gas price accordingly.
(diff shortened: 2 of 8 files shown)


local-test-configuration/airseeker/airseeker.example.json
```diff
@@ -16,7 +16,8 @@
         "sanitizationSamplingWindow": 900,
         "sanitizationPercentile": 80,
         "scalingWindow": 120,
-        "maxScalingMultiplier": 2
+        "maxScalingMultiplier": 2,
+        "sanitizationMultiplier": 2
```
Siegrift: Note, that for the actual values @bbenligiray mentioned using `50` as percentile and (I think) `3` for sanitization multiplier.

src/gas-price/gas-price.ts
```diff
@@ -125,10 +106,10 @@ export const getRecommendedGasPrice = (chainId: string, providerName: string, sp
     return null;
   }
 
-  const lastUpdateTimestamp = sponsorLastUpdateTimestamp[sponsorWalletAddress];
-  // Check if the next update is a retry of a pending transaction
-  if (lastUpdateTimestamp) {
-    const pendingPeriod = Math.floor(Date.now() / 1000) - lastUpdateTimestamp;
+  // Check if the next update is a retry of a pending transaction and scale the gas price accordingly.
+  let gasPriceToUse = multiplyBigNumber(latestGasPrice, recommendedGasPriceMultiplier);
```
bdrhn9, line 110: It's more clear to have the calculation that multiplies by `recommendedGasPriceMultiplier` be done here 👍🏼

src/gas-price/gas-price.ts
```diff
@@ -141,37 +122,38 @@ export const getRecommendedGasPrice = (chainId: string, providerName: string, sp
...
+  // If necessary, sanitize the gas price and log a warning because this should not happen under normal circumstances.
+  if (hasSufficientSanitizationData && gasPriceToUse > sanitizationGasPriceCap) {
     logger.warn('Sanitizing gas price.', {
-      gasPrice: latestGasPrice.toString(),
-      percentileGasPrice: percentileGasPrice.toString(),
+      gasPrice: gasPriceToUse.toString(),
+      sanitizationGasPriceCap: sanitizationGasPriceCap.toString(),
     });
-    return multiplyBigNumber(percentileGasPrice, recommendedGasPriceMultiplier);
+    return sanitizationGasPriceCap;
```
bdrhn9, line 155: Probably we should be more generous while configuring `sanitizationMultiplier` because now it is also capping the result from scaling.

bdrhn9 (approved):
Everything looks consistent with the issue and slack thread 🔥

src/gas-price/gas-price.ts
Siegrift: Yeah. Also note that it might be a good idea to set `recommendedGasPriceMultiplier` to `1` because now it's going to be the latest gas price.

This is something that should be also changed in the example files to make it more clear.

PR #263, opened 2024-04-23
Stagger Signed API calls

### LINKED ISSUE #255: Consider Signed data fetching improvements
Recently, there have been many alerts like [these](https://api3.grafana.net/goto/-FuS4QaSR?orgId=1). These all come from Airseeker `t0`. I assume this is because on the mainnets there are not so many active data feeds.

The reason this happens is likely because:
1. The calls are [not staggered](https://github.com/api3dao/airseeker-v2/blob/d247b4af10fe6f9a21992a09281e99f5c2579607/src/data-fetcher-loop/data-fetcher-loop.ts#L87) - Although it's not clear how to stagger these. Alternatively adding a small delay in between the calls would help the throughput.
2. The geographical distance to the Signed APIs - Nothing we can do about it.

Related to this issue:
1. We should log more context what fetch failed - e.g. the URL (or just API3/Nodary and airnode).
2. We are making the Signed API calls always in the same order. It's likely that the later calls suffer from these timeouts more so we will likely be missing data for some Airnodes more than the others.
3. Make the Airseeker Grafana alert more strict (it's currently relaxed [to avoid spamming the channel](https://api3workspace.slack.com/archives/C06R2JCFSS1/p1713371779761579?thread_ts=1713359154.955599&cid=C06R2JCFSS1)).

Siegrift:
Closes https://github.com/api3dao/airseeker-v2/issues/255

## Rationale

I was thinking about adding a static delay between the requests, but that is ad-hoc. Instead, I chose to follow the same principle that we do in the update feeds loop. The batches are staggered and [runUpdateFeeds](https://github.com/api3dao/airseeker-v2/blob/96924b62c87f0a2a7c6185206cea70fb47772b04/src/update-feeds-loops/update-feeds-loops.ts#L153) overlap (but the logic of processing the same feed at the same time is low). It works well in practice and staggering here is much simpler - we only make an HTTP call, verify the data and save it to state.

I was thinking whether to shuffle the URLs before fetching them as per:
> We are making the Signed API calls always in the same order. It's likely that the later calls suffer from these timeouts more so we will likely be missing data for some Airnodes more than the others.

but same order works much better with staggering.

### DIFF
--- a/src/data-fetcher-loop/data-fetcher-loop.ts
+++ b/src/data-fetcher-loop/data-fetcher-loop.ts
@@ -60,9 +60,13 @@ export const callSignedApi = async (url: string, timeout: number): Promise<Signe
     return null;
   }
 
-  const { data } = signedApiResponseSchema.parse(goAxiosCall.data?.data);
+  const parseResult = signedApiResponseSchema.safeParse(goAxiosCall.data?.data);
+  if (!parseResult.success) {
+    logger.warn('Failed to parse signed API response.', { url });
+    return null;
+  }
 
-  return Object.values(data);
+  return Object.values(parseResult.data.data);
 };
 
 export const runDataFetcher = async () => {
--- a/src/data-fetcher-loop/data-fetcher-loop.ts
+++ b/src/data-fetcher-loop/data-fetcher-loop.ts
@@ -56,7 +56,7 @@ export const callSignedApi = async (url: string, timeout: number): Promise<Signe
   );
 
   if (!goAxiosCall.success) {
-    logger.warn('Failed to fetch data from signed API.', parseAxiosError(goAxiosCall.error));
+    logger.warn('Failed to fetch data from signed API.', { url, ...parseAxiosError(goAxiosCall.error) });
     return null;
   }
 
--- a/src/data-fetcher-loop/data-fetcher-loop.ts
+++ b/src/data-fetcher-loop/data-fetcher-loop.ts
@@ -2,10 +2,10 @@ import { go } from '@api3/promise-utils';
 import axios, { type AxiosResponse, type AxiosError } from 'axios';
 import { pick, uniq } from 'lodash';
 
-import { HTTP_SIGNED_DATA_API_TIMEOUT_MULTIPLIER } from '../constants';
 import { logger } from '../logger';
 import { getState } from '../state';
 import { signedApiResponseSchema, type SignedData } from '../types';
+import { sleep } from '../utils';
 
 import { purgeOldSignedData, saveSignedData } from './signed-data-state';
 
@@ -67,8 +67,6 @@ export const callSignedApi = async (url: string, timeout: number): Promise<Signe
 
 export const runDataFetcher = async () => {
   return logger.runWithContext({ dataFetcherCoordinatorId: Date.now().toString() }, async () => {
-    logger.info('Running data fetcher.');
-
     const state = getState();
     const {
       config: { signedDataFetchInterval },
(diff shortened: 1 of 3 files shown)


src/data-fetcher-loop/data-fetcher-loop.ts
```diff
@@ -77,32 +75,33 @@ export const runDataFetcher = async () => {
...
+    logger.debug('Signed API URLs.', { signedApiUrls });
     const urls = uniq(
       Object.values(signedApiUrls)
         .map((urlsPerProvider) => Object.values(urlsPerProvider))
         .flat(2)
     );
 
+    const urlCount = urls.length;
+    const staggerTimeMs = signedDataFetchIntervalMs / urlCount;
+    logger.info('Fetching signed data', { urlCount, staggerTimeMs });
```
Siegrift, line 91: I'm really curious how many Signed API calls do we make, so I've replaced the nearly useless `Running data fetcher.` message with this.

src/data-fetcher-loop/data-fetcher-loop.ts
```diff
@@ -77,32 +75,33 @@ export const runDataFetcher = async () => {
...
-            if (!signedDataApiResponse) return;
-
-            for (const signedData of signedDataApiResponse) {
-              saveSignedData(signedData);
-            }
-          },
-          { totalTimeoutMs: signedDataFetchIntervalMs }
-        )
-      )
+      urls.map(async (url, index) => {
```
Siegrift, line 93: This should be non-throwing so no need to wrap in go anymore.

bdrhn9: Can following statement throw an error? If can, we also may want to wrap parsing operation be done via promise-utils.
https://github.com/api3dao/airseeker-v2/blob/96924b62c87f0a2a7c6185206cea70fb47772b04/src/data-fetcher-loop/data-fetcher-loop.ts#L63

src/data-fetcher-loop/data-fetcher-loop.ts
```diff
@@ -77,32 +75,33 @@ export const runDataFetcher = async () => {
...
+        await sleep(staggerTimeMs * index);
+
+        // NOTE: We allow each Signed API call to take full signedDataFetchIntervalMs. Because these calls are
+        // staggered, it means that there can be pending requests from different data fetcher loops happening at the
+        // same time. This does not matter much, because we only save the freshest signed data.
+        const signedDataApiResponse = await callSignedApi(url, signedDataFetchIntervalMs);
```
bdrhn9, line 99: If we decrease `signedDataFetchIntervalMs` to value like 2.5 or 5 secs, it can cause issues but otherwise (selecting it to be greater than 10 secs) it is feasible.

bdrhn9 (approved):
Looks good to me, I'm looking forward to deploy this version :)

src/data-fetcher-loop/data-fetcher-loop.ts
Siegrift: > Can following statement throw an error? If can, we also may want to wrap parsing operation be done via promise-utils.

Yeah. I previously though it can't because we fetch only from our Signed APIs, but later we might be fetching from the ones that are specified in the contract (which will be arbitrary) - so yeah, this should be wrapped.

src/data-fetcher-loop/data-fetcher-loop.ts
Siegrift: What is wrong with 2.5s ? I mean the timeout is probably a bit tight, but 5s should be quite enough when the calls are staggered imo.

bdrhn9: From my past experiences, 5 secs wasn't enough for calling airnode's signed data gateways. Signed API is more agile but it provides relatively big data (more than hundred signed data), so it's open to be time-out, if everything doesn't go perfect.

PR #264, opened 2024-04-23
Use logic from commons

### LINKED ISSUE #250: Use types from@api3/commons
[Here](https://github.com/api3dao/airseeker-v2/blob/d247b4af10fe6f9a21992a09281e99f5c2579607/src/types.ts#L5) we could use the types from commons instead. Also, some types are overly specific and they don't provide much value (e.g. `TemplateId` could be `Hex` or just a `string`).

Siegrift:
Closes https://github.com/api3dao/airseeker-v2/issues/250

## Rationale

While implementing the issue I realised there is some more code that should be reused from commons. Each commit focuses on a particular feature implemented in commons.


src/state/state.ts
```diff
@@ -19,16 +11,16 @@ interface GasPriceInfo {
 
 export interface State {
   config: Config;
-  gasPrices: Record<ChainId, Record<ProviderName, GasPriceInfo[]>>;
+  gasPrices: Record<string /* Chain ID */, Record<string /* Provider name */, GasPriceInfo[]>>;
```
bdrhn9: Can we use [this](https://github.com/api3dao/commons/blob/fb07e2c083cf63bc35d500aca37eb3fb7ee411c8/src/blockchain-utilities/schema.ts#L33) for `chainId`?

src/state/state.ts
```diff
@@ -19,16 +11,16 @@ interface GasPriceInfo {
...
-    ChainId,
-    Record<ProviderName, Record<Address /* Sponsor wallet */, number | null>>
+    string /* Chain ID */,
+    Record<string /* Provider name */, Record<Address /* Sponsor wallet */, number | null>>
   >;
-  derivedSponsorWallets: Record<DapiNameOrDataFeedId, PrivateKey>;
-  signedDatas: Record<BeaconId, SignedData>;
-  signedApiUrls: Record<ChainId, Record<ProviderName, SignedApiUrl[]>>;
+  derivedSponsorWallets: Record<string /* dAPI name or data feed ID */, Hex /* Private key */>;
+  signedDatas: Record<Hex /* Beacon ID */, SignedData>;
```
bdrhn9, line 22: [Keccak256Hash](https://github.com/api3dao/commons/blob/fb07e2c083cf63bc35d500aca37eb3fb7ee411c8/src/blockchain-utilities/schema.ts#L26C13-L26C26) for beacon Ids?

src/types.ts
```diff
@@ -1,20 +1,10 @@
...
-export type DapiNameOrDataFeedId = string;
-
 // Taken from https://github.com/api3dao/signed-api/blob/main/packages/api/src/schema.ts
 export const signedDataSchema = z.object({
-  airnode: evmAddressSchema,
-  templateId: evmIdSchema,
+  airnode: addressSchema,
+  templateId: keccak256HashSchema,
   timestamp: z.string(),
   encodedValue: z.string(),
```
bdrhn9: [hexSchema](https://github.com/api3dao/commons/blob/fb07e2c083cf63bc35d500aca37eb3fb7ee411c8/src/blockchain-utilities/schema.ts#L5) for `encodedValue` and `signature`?

src/update-feeds-loops/contracts.ts
```diff
@@ -51,9 +51,9 @@ export const decodeGetBlockNumberResponse = Number;
 export const decodeGetChainIdResponse = Number;
 
 export interface Beacon {
-  airnodeAddress: AirnodeAddress;
-  templateId: TemplateId;
-  beaconId: string;
+  airnodeAddress: Address;
+  templateId: Hex;
```
bdrhn9, line 55: [Keccak256Hash](https://github.com/api3dao/commons/blob/fb07e2c083cf63bc35d500aca37eb3fb7ee411c8/src/blockchain-utilities/schema.ts#L26C13-L26C26) for beacon and template Ids?

bdrhn9 (approved):
Minor suggestions other than LGTM.

src/state/state.ts
Siegrift: TBH I am not a great fan of that type. It's quite verbose to read/type and I don't see the benefit compared to `Hex`. I'd just stick with `Hex` unless there is a better reason for swithcing.

Also, TS will treat the `Hex` and `Keccak256Hash` as the same - so in the future if these are mixed it may cause unnecessary confusion.

src/update-feeds-loops/contracts.ts
Siegrift: See: https://github.com/api3dao/airseeker-v2/pull/264#discussion_r1577830760

PR #265, opened 2024-04-24
Minor log improvements

### LINKED ISSUE #244: Improve gas sanitization/scaling logs & alerts
Currently, we log a warning when gas scaling and sanitization happens, but this is a quite frequent event and we don't have capacity to inspect each occurrence. Also, often the gas scaling/sanitization is a one time thing which is acceptable.

In contrast, there if the transaction is pending for a long time, then it's something serious. Similarly, if we sanitize repeatedly or the sanitized gas price differs by a large percentage. 

I suggest:
- [x] Log how many times a particular tx has been scaled. Note, that we already log a pending period - but this depends on the scaling window and how data fetching interval. E.g. the tx can be pending for 30 seconds before it is scaled for the first time. We could also log how long is the feed pending as a percentage of scaling window. The alert should warn when the tx needs to be scaled repeatedly.
- [x] Log sanitization percentage (sanitized price vs RPC price)

### LINKED ISSUE #259: Log improvements after alert investigation
* https://github.com/api3dao/airseeker-v2/blob/main/src/update-feeds-loops/submit-transactions.ts#L105 should be `sponsorWalletAddress`
* When submitting tx, log it's tx ID - This can help tracking whether the TX went through or not. Otherwise one needs to look at all transactions made by sponsor wallet around the specific time.
* When clearing timestamp, also note how long it took until it was cleared (now - firstDeviationExceeded).

Siegrift:
Closes https://github.com/api3dao/airseeker-v2/issues/259
Closes https://github.com/api3dao/airseeker-v2/issues/244

### DIFF
--- a/src/gas-price/gas-price.test.ts
+++ b/src/gas-price/gas-price.test.ts
@@ -160,6 +160,7 @@ describe(getRecommendedGasPrice.name, () => {
     expect(logger.warn).toHaveBeenCalledTimes(1);
     expect(logger.warn).toHaveBeenNthCalledWith(1, 'Sanitizing gas price.', {
       gasPrice: '120000000000',
+      ratio: '6.98',
       sanitizationGasPriceCap: '17200000000',
     });
   });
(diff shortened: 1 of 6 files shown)


bdrhn9 (approved):
LGTM 👍🏼

PR #269, opened 2024-04-25
Update renovate dependency reviewer

Siegrift:
I think it makes more sense for me to handle the dependency updates - it's basically 0 effort anyway. I'll update the responsibilities in our tech documentation afterwards.


bdrhn9 (approved):
It makes sense.

PR #286, opened 2024-04-30
Fix gas price scaling

### LINKED ISSUE #284: Only scale gas price when the transaction is a retry
I've noticed that we always scale the gas price (even when the transaction is the original). This is because we set the pending timestamp before making the transaction (so during update Airseeker already sees the transaction as a retry).

[This code](https://github.com/api3dao/airseeker-v2/blob/451e8c9ee53bcb969b5e1156de196a43d0a418dc/src/update-feeds-loops/update-feeds-loops.ts#L319) needs to be revisited.

Siegrift:
Closes https://github.com/api3dao/airseeker-v2/issues/284

## Rationale

We are applying the pending timestamp before we submit the transaction. This breaks the check for pending transactions - making it seem that every transaction is a retry causing Airseeker to scale it. This patch adds a test ensuring this won't happen again.

At first, I wanted to move the logic after the transactions are submitted, but this is awkward for two reasons:
1. The pending timestamp needs to be applied even if the submission fails (or unexpectedly throws) - making this more verbose to implement correctly.
2. The timestamp should be applied at the update condition check (not after we tried to submit update txs). Again, making it more awkward to implement.

For the above reasons, I applied the logic for tracking pending transactions immediately after checking the update conditions, but added a new field `consecutiveUpdates` which can be used to check how many times this feed was updatable.


bdrhn9 (approved):
LGTM 👍🏼

Siegrift:
FYI: I've confirmed that the `v3.0.3` does not scale the gas price for original transactions.

PR #291, opened 2024-05-02
Verify signed data using worker thread

### LINKED ISSUE #282: Utilize worker threads
@bdrhn9 had a good idea that by spawning multiple workers we might leverage the additional CPU cores. This would also reduce the overhead of the "main" loop that currently does everything.

From the quick look there don't seem to be any downsides to implementing this.

Siegrift:
Closes https://github.com/api3dao/airseeker-v2/issues/282

## Rationale

Apply the same fix as we did for Signed APIs. This has proven to be enough. Another optimization on top of this is to validate only the signed data for active beacons (aka - beacons of active feeds). I needed to refactor the code a bit to accommodate for the changes.

### DIFF
--- a/src/data-fetcher-loop/signed-data-state.ts
+++ b/src/data-fetcher-loop/signed-data-state.ts
@@ -60,7 +60,7 @@ export const saveSignedData = async (signedDataBatch: SignedDataRecordEntry[]) =
   return signedDataBatch.length;
 };
 
-export const getSignedData = (dataFeedId: Hex) => getState().signedDatas[dataFeedId];
+export const getSignedData = (beaconId: Hex) => getState().signedDatas[beaconId];
 
 export const isSignedDataFresh = (signedData: SignedData) =>
   BigInt(signedData.timestamp) > BigInt(Math.ceil(Date.now() / 1000 - 24 * 60 * 60));
@@ -74,7 +74,7 @@ export const purgeOldSignedData = () => {
 
   updateState((draft) => {
     draft.signedDatas = Object.fromEntries(
-      Object.entries(draft.signedDatas).filter(([_dataFeedId, signedData]) => isSignedDataFresh(signedData!))
+      Object.entries(draft.signedDatas).filter(([_beaconId, signedData]) => isSignedDataFresh(signedData!))
     );
   });
 };
--- a/src/data-fetcher-loop/signed-data-state.ts
+++ b/src/data-fetcher-loop/signed-data-state.ts
@@ -8,7 +8,7 @@ import type { SignedData } from '../types';
 
 export const verifySignedData = ({ airnode, templateId, timestamp, signature, encodedValue }: SignedData) => {
   // Verification is wrapped in goSync, because ethers methods can potentially throw on invalid input.
-  const goVerify = goSync(() => {
+  const goVerifySignature = goSync(() => {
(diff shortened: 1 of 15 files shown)


src/data-fetcher-loop/signed-data-state.ts
```diff
@@ -1,100 +1,80 @@
...
-    const state = getState();
-
-    const dataFeedId = deriveBeaconId(airnode, templateId) as Hex;
-
-    const existingValue = state.signedDatas[dataFeedId];
-    if (existingValue && existingValue.timestamp >= timestamp) {
-      logger.debug('Skipping state update. The signed data value is not fresher than the stored value.');
-      return;
+  const verifier = await getVerifier();
+  const verificationResult = await verifier.verifySignedData(signedDataBatch);
```
bdrhn9, line 49: We skipping whole batch if one of them is defective. However, it looks okay for me, we don't expect any faulty signed data.

src/update-feeds-loops/update-feeds-loops.ts
```diff
@@ -251,12 +251,18 @@ export const runUpdateFeeds = async (providerName: string, chain: Chain, chainId
...
         );
-        // Overwrite the state with the new signed API URLs instead of merging them to avoid stale URLs.
-        updateState((draft) => set(draft, ['signedApiUrls', chainId, providerName], signedApiUrls));
+        const beaconIds = uniq(
+          processedBatches.reduce<string[]>((acc, batch) => (batch ? [...acc, ...batch.beaconIds] : acc), [])
+        );
+        // Overwrite the state with the new signed API URLs instead of merging them to avoid keeping stale URLs.
+        updateState((draft) => {
+          set(draft, ['signedApiUrls', chainId, providerName], signedApiUrls);
+          set(draft, ['beaconIds', chainId, providerName], beaconIds);
```
bdrhn9:
```suggestion
          set(draft, ['activeDataFeedBeaconIds', chainId, providerName], beaconIds);
```

src/data-fetcher-loop/signed-data-state.ts
```diff
@@ -1,100 +1,80 @@
...
+  if (verificationResult !== true) {
```
bdrhn9, line 52: We skipping whole batch if one of them is defective. However, it looks okay for me, we don't expect any faulty signed data.

bdrhn9 (commented):
I think these changes will help CPU to distribute the load evenly 👍🏼

src/data-fetcher-loop/signed-data-state.ts
Siegrift: Yeah, that's questionable part, but it felt natural to me because we are doing the same in the Signed API https://github.com/api3dao/signed-api/blob/main/packages/signed-api/README.md#api

I'll leave a comment about it though.

src/update-feeds-loops/update-feeds-loops.ts
Siegrift: Btw. this is why I am not a fan of these lodash utils - they are unable to infer the paths (even though they could).

Siegrift:
I'm going to implement the config Signed API whitelisting separately. See: https://github.com/api3dao/airseeker-v2/issues/300

bdrhn9 (approved):

PR #292, opened 2024-05-02
Change coordinator IDs to ISO8601 format

### LINKED ISSUE #280: Use `uuid` as coordinator ID
Current timestamp is utilized as coordinator ID for both `dataFetcherCoordinatorId` and `updateFeedsCoordinatorId`. However timestamp isn't good choice to discriminate invoked functions especially there is bug related to race conditioning. That being said, because assigning coordinator ID is being done before staggering, coordinator IDs for same data feed from different RPC providers results in same.

Siegrift:
Closes https://github.com/api3dao/airseeker-v2/issues/280

## Rationale

This should make the IDs easy to differentiate from each other, while being "unique enough". I don't see a huge need for UUIDs.

### DIFF
--- a/src/data-fetcher-loop/data-fetcher-loop.ts
+++ b/src/data-fetcher-loop/data-fetcher-loop.ts
@@ -4,7 +4,7 @@ import { uniq } from 'lodash';
 import { logger } from '../logger';
 import { getState } from '../state';
 import { type SignedDataRecord, signedApiResponseSchema, type SignedDataRecordEntry } from '../types';
-import { sleep } from '../utils';
+import { generateRandomId, sleep } from '../utils';
 
 import { purgeOldSignedData, saveSignedData } from './signed-data-state';
 
@@ -56,7 +56,7 @@ export const callSignedApi = async (url: string, timeout: number): Promise<Signe
 };
 
 export const runDataFetcher = async () => {
-  return logger.runWithContext({ dataFetcherCoordinatorId: new Date().toISOString() }, async () => {
+  return logger.runWithContext({ dataFetcherCoordinatorId: generateRandomId() }, async () => {
     const state = getState();
     const {
       config: { signedDataFetchInterval },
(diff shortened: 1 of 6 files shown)


src/data-fetcher-loop/data-fetcher-loop.ts
```diff
@@ -56,7 +56,7 @@ export const callSignedApi = async (url: string, timeout: number): Promise<Signe
 };
 
 export const runDataFetcher = async () => {
-  return logger.runWithContext({ dataFetcherCoordinatorId: Date.now().toString() }, async () => {
+  return logger.runWithContext({ dataFetcherCoordinatorId: new Date().toISOString() }, async () => {
```
bdrhn9: If we continue with time, I'd prefer timestamp rather than ISO string cuz it's easier to search logs with timestamp. When you double-click on timestamp it will select all, it's not possible with ISO format 😄

Siegrift: :D UUID would have the same issue though.

Anyway, let's close this and stick with timestamp.

bdrhn9: The objective of switching to `uuid` is not about the double-click, it guarantees that the uniqueness. And it doesn't have to be `uuid` it can be 32-bit hex generated via `node:crypto` library.

Siegrift: Yeah, but the JS timestamp is in MS so that's very low probability that the loops start at the same time. Anyway, 32-bit hex from node:crypto + a timestamp logged at the start of the coordinator seems like a trade-off that everyone is going to be happy with. WDYT?

bdrhn9: Yes, it looks like sweet-spot.

src/data-fetcher-loop/data-fetcher-loop.ts
```diff
@@ -83,7 +83,7 @@ export const runDataFetcher = async () => {
 
     const urlCount = urls.length;
     const staggerTimeMs = signedDataFetchIntervalMs / urlCount;
-    logger.info('Fetching signed data.', { urlCount, staggerTimeMs });
+    logger.info('Fetching signed data.', { urlCount, staggerTimeMs, currentTime: new Date().toISOString() });
```
Siegrift, line 86: This message is logged as a basically the first thing, so I've logged the current time here instead of logging another message.

PR #294, opened 2024-05-02
Fix ethers error messages sanitization

### LINKED ISSUE #276: Trim ethers errors
See [this log](https://api3.grafana.net/goto/7s7wnUfSR?orgId=1). It's too large.

The error message should be trimmed.

Siegrift:
Closes https://github.com/api3dao/airseeker-v2/issues/276

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -91,7 +91,12 @@ export const sanitizeEthersError = (error: Error) => {
 
   // We don't care about the stack trace, nor error name - just the code and the message. According to the ethers
   // sources, the short message should always be defined.
-  return new SanitizedErrorsError(ethersError.code, ethersError.shortMessage);
+  const sanitizedError = new SanitizedErrorsError(ethersError.code, ethersError.shortMessage);
+  // NOTE: We don't need the stack trace, because the errors are usually easy to find by the developer message and the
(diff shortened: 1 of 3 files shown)


src/utils.ts
```diff
@@ -89,5 +89,10 @@ export const sanitizeEthersError = (error: Error) => {
 
   // We don't care about the stack trace, nor error name - just the code and the message. According to the ethers
   // sources, the short message should always be defined.
-  return new SanitizedErrorsError(ethersError.code, ethersError.shortMessage);
+  const sanitizedError = new SanitizedErrorsError(ethersError.code, ethersError.shortMessage);
+  // NOTE: We don't need the stack trace, because the errors are usually easy to find by the developer message and the
+  // stack can be traced manually. This reduces the risk of the stack trace being too large and "exploding" the log
+  // size.
+  delete sanitizedError.stack;
```
bdrhn9, line 98: We didn't pass `stack` attribute of `error`. Do we still need to delete it?

bdrhn9 (approved):

src/utils.ts
Siegrift: Yeah, because every `Error` (even if extended) has one.

PR #296, opened 2024-05-02
Refactor to allow for single sponsorWallet updates 

acenolaza:
### What does this change?

It basically applies suggestion made [here](https://api3workspace.slack.com/archives/C05TQPT7PNJ/p1714555878380499?thread_ts=1714471146.477969&cid=C05TQPT7PNJ) to allow sending txs using a single sponsor wallet. This is meant to be used by a deployment used a fallback that should send txs whenever main airseeker deployments are down their sponsorWallets have run out of funds for some reason. This fallback deployment also makes our lives easier by not having to top-up the sponsorWallet of each dAPI.

1. Adds a new wallet derivation scheme type called `fallback`
2. It also adds a new property to `walletDerivationScheme` config object called `sponsorAddress`. This property must be set to a valid address when `walletDerivationScheme.type` is `fallback`
3. `pendingTransactionsInfo` objects are now keyed by `dataFeedId` instead of `sponsorWalletAddress`
4. The `getRecommendedGasPrice()` nows receives an array of `dataFeedIds` instead of a single `sponsorWalletAddress` as argument and it only uses the oldest `pendingTransactionsInfo` (when more than 1 `dataFeedId` is passed)
5. There is a new `submitBatchTransaction()` function that bundles all the calldatas required to update all data feeds in the batch and sends a single `tryMulticall()` tx.


bbenligiray:
> It also adds a new property to `walletDerivationScheme` config object called `sponsorWallet`. 

Looks like you mean `sponsorAddress`.

I dislike "`fallback`", something like "`fixed`" would be semantically less restrictive

> pendingTransactionsInfo objects are now keyed by dataFeedId instead of sponsorWalletAddress

I thought we decided not to touch this on a recent call. Doesn't self-funded scheme update the same data feed (with ID) using multiple sponsor wallets belonging to different update parameters?

acenolaza:
> Looks like you mean `sponsorAddress`.

Good catch. Updated comment

> I dislike "`fallback`", something like "`fixed`" would be semantically less restrictive

👍🏻 

> I thought we decided not to touch this on a recent call.

That is correct but Emanuel made a good suggestion where by doing this change, retries get mostly fixed for single wallet

> Doesn't self-funded scheme update the same data feed (with ID) using multiple sponsor wallets belonging to different update parameters?

🤔 that is a good question and it seems we could be scaling the gas price for wallets that might not need it when self-funded is set. @Siegrift any thoughts on this? Maybe we can use both as keys 🤔 

```
pendingTransactionsInfo: Record<
    ChainId,
    Record<string /* Provider name */, Record<Address /* Sponsor wallet */, Record<Hex /* Data Feed ID */, PendingTransactionInfo | null>>>
  >;
```

acenolaza:
> Maybe we can use both as keys 🤔

Implemented that [here](https://github.com/api3dao/airseeker-v2/pull/296/commits/96937b5fa1b525df3d5e755fadc8bc5be16c7f3b) I'll wait for comments 🙏🏻

Siegrift (commented):
> > Doesn't self-funded scheme update the same data feed (with ID) using multiple sponsor wallets belonging to different update parameters?

> 🤔 that is a good question and it seems we could be scaling the gas price for wallets that might not need it when self-funded is set. @Siegrift any thoughts on this? Maybe we can use both as keys 🤔

Currently, Airseeker can update the same data feed using different wallets for self-funded (as @bbenligiray suggests in the question), but it can also happen when there are 2 active data feed using the same data feed ID (e.g. one identified by ID, one by dapiName). Airseeker now treats these separate and scales them independently. There is no case, where a single wallet updates multiple data feeds, so key-ing by both data feed ID and sponsor address solves the problem for "fixed" wallet derivation.

Another question is whether it doesn't make sense for this to only key this by data feed ID. In both managed/self-funded there we will be scaling unnecessarily (in the self-funded it depends on how often the deviation is triggered for multiple deviating feeds - but that seems probably enough). So I think it doesn't and we should do what you suggest.

> I dislike "fallback", something like "fixed" would be semantically less restrictive

fwiw: I don't have a strong opinion about this.

src/gas-price/gas-price.ts
```diff
@@ -81,9 +81,24 @@ export const fetchAndStoreGasPrice = async (
...
+  const pendingTransactionInfos = dataFeedIds.reduce((acc: PendingTransactionInfo[], dataFeedId) => {
+    const pendingTransactionInfo =
+      state.pendingTransactionsInfo[chainId]![providerName]![sponsorWalletAddress]![dataFeedId];
+    if (!pendingTransactionInfo) {
+      return acc;
+    }
+    return [...acc, pendingTransactionInfo];
+  }, []);
+  // Get the oldest PendingTransactionInfo
+  const pendingTransactionInfo = minBy(pendingTransactionInfos, (info) => info?.firstUpdatableTimestamp);
```
Siegrift: You can do it without the reduce above.

src/state/state.ts
```diff
@@ -23,7 +23,10 @@ export interface State {
   gasPrices: Record<ChainId, Record<string /* Provider name */, GasPriceInfo[]>>;
   pendingTransactionsInfo: Record<
     ChainId,
-    Record<string /* Provider name */, Record<Address /* Sponsor wallet */, PendingTransactionInfo | null>>
+    Record<
+      string /* Provider name */,
+      Record<Address /* Sponsor wallet */, Record<Hex /* Data Feed ID */, PendingTransactionInfo | null>>
```
Siegrift, line 33: This deserves an inline comment + spec comment.

src/update-feeds-loops/submit-transactions.ts
Siegrift, line 1: It looks like this module copy-pastes a lot of existing functionality. Also, some changes are unnecessary and can be reverted.

README.md
```diff
@@ -273,6 +273,8 @@ The following options are available:
   parameters. This is the scheme that was originally used by Nodary for self-funded data feeds.
 - `managed` - Derives the wallet from the hash of the dAPI name (or data feed ID). This means the wallet derivation is
   agnostic to update parameters, and the same wallet is used when the dAPI is upgraded/downgraded.
+- `fixed` - Derives the wallet from the hash configured in a sibling property called `sponsorAddress`. This means the
```
Siegrift:
```suggestion
- `fixed` - Derives the wallet from the specified `sponsorAddress`. All data feed updates will be done via this single wallet.
```

local-test-configuration/scripts/initialize-chain.ts
```diff
@@ -57,24 +57,55 @@ export const deriveRole = (adminRole: string, roleDescription: string) => {
...
   // Initialize sponsor wallets
-  for (const beaconSetName of getBeaconSetNames()) {
+  const beaconSetNames = await getBeaconSetNames();
+  for (const beaconSetName of beaconSetNames) {
     const dapiName = encodeDapiName(beaconSetName);
-
-    const sponsorAddressHash = deriveSponsorAddressHashForManagedFeed(dapiName);
+    const updateParameters = encodeUpdateParameters();
+
+    let sponsorAddressHash;
```
Siegrift: TODO: mirror impl.

local-test-configuration/scripts/initialize-chain.ts
```diff
@@ -112,64 +143,92 @@ const loadAirnodeFeedConfig = (airnodeFeedDir: 'airnode-feed-1' | 'airnode-feed-
...
+    switch (rawConfig.walletDerivationScheme.type) {
+      case 'self-funded': {
+        sponsorAddressHash = deriveSponsorAddressHashForSelfFundedFeed(dapiName, updateParameters);
+        break;
+      }
+      case 'managed': {
+        sponsorAddressHash = deriveSponsorAddressHashForManagedFeed(dapiName);
+        break;
+      }
+      case 'fixed': {
```
Siegrift: TODO: Mirror impl.

local-test-configuration/scripts/initialize-chain.ts
```diff
@@ -112,64 +143,92 @@ const loadAirnodeFeedConfig = (airnodeFeedDir: 'airnode-feed-1' | 'airnode-feed-
...
+  const randomPerson = ethers.Wallet.createRandom().connect(deployerAndManager.provider);
+  const fundRandomPersonTx = await deployerAndManager.sendTransaction({
+    to: await randomPerson.getAddress(),
+    value: ethers.parseEther('1'),
+  });
+  await fundRandomPersonTx.wait();
 
   // Deploy contracts
-  const accessControlRegistryFactory = new AccessControlRegistryFactory(deployerAndManager as Signer);
+  const accessControlRegistryFactory = new AccessControlRegistryFactory(deployerAndManager);
```
Siegrift, line 185: 👍

local-test-configuration/scripts/initialize-chain.ts
```diff
@@ -112,64 +143,92 @@ const loadAirnodeFeedConfig = (airnodeFeedDir: 'airnode-feed-1' | 'airnode-feed-
...
   const accessControlRegistry = await accessControlRegistryFactory.deploy();
   await accessControlRegistry.waitForDeployment();
-  const api3ServerV1Factory = new Api3ServerV1Factory(deployerAndManager as Signer);
+  const api3ServerV1Factory = new Api3ServerV1Factory(deployerAndManager);
   const api3ServerV1AdminRoleDescription = 'Api3ServerV1 admin';
   const api3ServerV1 = await api3ServerV1Factory.deploy(
-    accessControlRegistry.getAddress(),
+    await accessControlRegistry.getAddress(),
```
Siegrift: It this necessary? Afaik ethers accepted a promise here as well.

src/config/schema.test.ts
```diff
@@ -202,3 +207,51 @@ describe('chains schema', () => {
     );
   });
 });
+
+describe('walletDerivationScheme schema', () => {
```
Siegrift, line 211: 👍

src/config/schema.ts
```diff
@@ -154,15 +154,20 @@ export const deviationThresholdCoefficientSchema = z
...
 
-export const walletDerivationTypeSchema = z.union([z.literal('self-funded'), z.literal('managed')]);
+export const walletDerivationTypeSchema = z.union([z.literal('self-funded'), z.literal('managed'), z.literal('fixed')]);
 
 export type WalletDerivationType = z.infer<typeof walletDerivationTypeSchema>;
 
 export const walletDerivationSchemeSchema = z
   .object({
     type: walletDerivationTypeSchema,
+    sponsorAddress: addressSchema.optional(),
```
Siegrift: This makes it possible to supply the `sponsorAddress` on other derivation schemes, which we don't want. Do this instead:
```ts
export const walletDerivationSchemeSchema = z.discriminatedUnion('type', [
  z.object({ type: z.literal('self-funded') }),
  z.object({ type: z.literal('managed') }),
  z.object({ type: z.literal('fixed'), sponsorAddress: addressSchema }),
]);

export type WalletDerivationScheme = z.infer<typeof walletDerivationSchemeSchema>;

export type WalletDerivationType = WalletDerivationScheme['type'];
```

src/gas-price/gas-price.test.ts
```diff
@@ -312,6 +326,51 @@ describe(getRecommendedGasPrice.name, () => {
       sanitizationGasPriceCap: '19400000000',
     });
   });
+
+  it('applies scaling based on oldest non-null PendingTransactionInfo when multiple dataFeedIds are passed', () => {
```
Siegrift:
```suggestion
  it('applies scaling based on oldest non-null pending data feed when multiple data feed IDs are passed', () => {
```

src/gas-price/gas-price.ts
```diff
@@ -81,9 +81,27 @@ export const fetchAndStoreGasPrice = async (
...
-export const getRecommendedGasPrice = (chainId: string, providerName: string, sponsorWalletAddress: Address) => {
+export const getRecommendedGasPrice = (
+  chainId: string,
+  providerName: string,
+  sponsorWalletAddress: Address,
+  dataFeedIds: Hex[]
+) => {
   const state = getState();
-  const pendingTransactionInfo = state.pendingTransactionsInfo[chainId]![providerName]![sponsorWalletAddress];
+  const pendingTransactionInfos = dataFeedIds.reduce((acc: PendingTransactionInfo[], dataFeedId) => {
```
Siegrift: This would be much more readable with `.map(<select_pending_tx_info>).filter(Boolean)`

src/gas-price/gas-price.ts
```diff
@@ -81,9 +81,27 @@ export const fetchAndStoreGasPrice = async (
...
+    const pendingTransactionInfo =
+      state.pendingTransactionsInfo[chainId]![providerName]![sponsorWalletAddress]?.[dataFeedId];
+    if (!pendingTransactionInfo) {
+      return acc;
+    }
+    return [...acc, pendingTransactionInfo];
+  }, []);
+  // Get the oldest PendingTransactionInfo and if the oldest is not a single object then minBy will return the one with the largest consecutivelyUpdatableCount.
```
Siegrift: Do not refer to interface/variable names because they go stale quickly unless absolutily necessary.

src/gas-price/gas-price.ts
```diff
@@ -81,9 +81,27 @@ export const fetchAndStoreGasPrice = async (
...
+  const pendingTransactionInfo = minBy(
```
Siegrift: I'd rename to `oldestPendingTransactionInfo`

src/update-feeds-loops/submit-transactions.test.ts
```diff
@@ -305,6 +310,177 @@ describe(submitTransactionsModule.submitTransactions.name, () => {
 
     expect(submitTransactionsModule.submitTransaction).toHaveBeenCalledTimes(2);
   });
+
+  it('submits a single transaction for updating all feeds in the batch', async () => {
```
Siegrift, line 320: 👍

src/update-feeds-loops/submit-transactions.test.ts
```diff
@@ -305,6 +310,177 @@ describe(submitTransactionsModule.submitTransactions.name, () => {
...
+    expect(logger.debug).toHaveBeenNthCalledWith(2, 'Derived new sponsor wallet.', {
+      sponsorWalletAddress: '0xFaFF9C2E67716d2209552f46Fa9829D46830aCcB',
+    });
+    expect(logger.debug).toHaveBeenNthCalledWith(3, 'Getting nonce.');
+    expect(logger.debug).toHaveBeenNthCalledWith(4, 'Getting recommended gas price.');
+    expect(logger.debug).toHaveBeenNthCalledWith(5, 'Creating calldatas.');
+    expect(logger.debug).toHaveBeenNthCalledWith(6, 'Estimating batch update gas limit.');
+  });
+
+  it('logs and error when getting nonce fails', async () => {
```
Siegrift:
```suggestion
  it('logs an error when getting nonce fails', async () => {
```

src/update-feeds-loops/submit-transactions.test.ts
```diff
@@ -317,7 +493,7 @@ describe(submitTransactionsModule.submitTransaction.name, () => {
     jest.spyOn(gasEstimationModule, 'estimateMulticallGasLimit').mockResolvedValue(BigInt(500_000));
     jest.spyOn(gasPriceModule, 'getRecommendedGasPrice').mockReturnValue(BigInt(100_000_000));
     const api3ServerV1 = generateMockApi3ServerV1();
-    jest.spyOn(api3ServerV1.tryMulticall, 'send').mockReturnValue({ hash: '0xTransactionHash' });
+    jest.spyOn(api3ServerV1, 'tryMulticall').mockReturnValue({ hash: '0xTransactionHash' });
```
Siegrift: Any reason to change this from `.send` to a function call?

src/update-feeds-loops/submit-transactions.ts
```diff
@@ -42,15 +43,15 @@ export const submitUpdate = async (
...
+  sponsorWallet: ethers.HDNodeWallet | ethers.Wallet,
   gasPrice: bigint,
   nonce: number
 ) => {
   const {
     updatableBeacons,
     dataFeedInfo: { beaconsWithData },
   } = updatableDataFeed;
-  const sponsorWalletAddress = connectedSponsorWallet.address as Address;
+  const sponsorWalletAddress = (await sponsorWallet.getAddress()) as Address;
```
Siegrift: Why this change?

src/update-feeds-loops/submit-transactions.ts
```diff
@@ -70,8 +71,8 @@ export const submitUpdate = async (
       signedData: { airnode, templateId, timestamp, encodedValue, signature },
     } = beacon;
     return api3ServerV1
-      .connect(connectedSponsorWallet)
-      .updateBeaconWithSignedData.send(airnode, templateId, timestamp, encodedValue, signature, {
+      .connect(sponsorWallet)
+      .updateBeaconWithSignedData(airnode, templateId, timestamp, encodedValue, signature, {
```
Siegrift: Why this change?

src/update-feeds-loops/submit-transactions.ts
```diff
@@ -91,9 +92,120 @@ export const submitUpdate = async (
...
+  provider: ethers.JsonRpcProvider,
+  api3ServerV1: Api3ServerV1,
+  updatableDataFeeds: UpdatableDataFeed[],
+  blockNumber: number
+) => {
+  const {
+    config: { chains, sponsorWalletMnemonic, walletDerivationScheme },
+  } = getState();
+
+  const [decodedDapiNames, dataFeedIds] = updatableDataFeeds.reduce(
```
Siegrift: This is too unreadable. Just do a two `.map` calls instead.

src/update-feeds-loops/submit-transactions.ts
```diff
@@ -91,9 +92,120 @@ export const submitUpdate = async (
...
+  return logger.runWithContext({ dapiNames: decodedDapiNames, dataFeedIds, blockNumber }, async () => {
+    // NOTE: We use go mainly to set a timeout for the whole update process. We expect the function not to throw and
+    // handle errors internally.
+    const goUpdate = await go(
+      async () => {
+        logger.debug('Getting derived sponsor wallet.');
+        const sponsorWallet = getDerivedSponsorWallet(
+          sponsorWalletMnemonic,
+          ethers.ZeroHash as Hex,
+          '',
```
Siegrift: It feels like this function should better define the input parameters. Specifically, the update parameters here are only needed for self funded feeds.

Alternatively, add a comment here to explain why we pass `''` (it's unused).

src/update-feeds-loops/submit-transactions.ts
```diff
@@ -91,9 +92,120 @@ export const submitUpdate = async (
...
+        [...dataFeedIds, dataFeedId],
+      ];
+    },
+    [[], []]
+  );
+
+  const { dataFeedUpdateInterval, fallbackGasLimit } = chains[chainId]!;
+  const dataFeedUpdateIntervalMs = dataFeedUpdateInterval * 1000;
+
+  return logger.runWithContext({ dapiNames: decodedDapiNames, dataFeedIds, blockNumber }, async () => {
```
Siegrift, line 146: This is too similar with `submitTransaction`. I think we can do:
1. In `submitUpdate` change `updatableDataFeed` to `updatableDataFeeds` array.
2. Wrap `submitUpdate` in go and handle [this](https://github.com/api3dao/airseeker-v2/blob/35aa3a66a7bf25f8983032e83aaaa6a7216ea4d4/src/update-feeds-loops/submit-transactions.ts#L258) inside.

src/update-feeds-loops/submit-transactions.ts
```diff
@@ -125,8 +237,8 @@ export const submitTransaction = async (
           dapiName ?? dataFeedId,
           updateParameters,
           walletDerivationScheme
-        );
-        const sponsorWalletAddress = sponsorWallet.address as Address;
+        ).connect(provider);
```
Siegrift: Move this comment:
```
          // When we add the sponsor wallet (signer) without connecting it to the provider, the provider of the
          // contract will be set to "null". We need to connect the sponsor wallet to the provider of the contract.
```

here.

src/update-feeds-loops/submit-transactions.ts
```diff
@@ -125,8 +237,8 @@ export const submitTransaction = async (
...
+        const sponsorWalletAddress = (await sponsorWallet.getAddress()) as Address;
```
Siegrift: Why this change?

src/update-feeds-loops/submit-transactions.ts
```diff
@@ -196,12 +298,25 @@ export const submitTransactions = async (
...
-    updatableDataFeeds.map(async (dataFeed) =>
-      submitTransaction(chainId, providerName, provider, api3ServerV1, dataFeed, blockNumber)
-    )
-  );
+) => {
+  const {
+    config: { walletDerivationScheme },
+  } = getState();
+  if (walletDerivationScheme.type === 'fixed') {
+    if (isEmpty(updatableDataFeeds)) {
```
Siegrift: Imo, this should be handled in `submitBatchTransaction`. Also, make it a single line return.

src/update-feeds-loops/submit-transactions.ts
```diff
@@ -196,12 +298,25 @@ export const submitTransactions = async (
...
+      return [];
+    }
+    return Array.from({ length: updatableDataFeeds.length }).fill(
+      await submitBatchTransaction(chainId, providerName, provider, api3ServerV1, updatableDataFeeds, blockNumber)
```
Siegrift: This should be called only once (for all updatable feeds).

src/update-feeds-loops/submit-transactions.ts
```diff
@@ -196,12 +298,25 @@ export const submitTransactions = async (
...
+    );
+  } else {
```
Siegrift: Omit the else part in favour "if-return" pattern.

src/utils.ts
```diff
@@ -47,10 +47,21 @@ export const deriveSponsorWallet = (
...
+    case 'self-funded': {
+      sponsorAddressHash = deriveSponsorAddressHashForSelfFundedFeed(dapiNameOrDataFeedId, updateParameters);
+      break;
+    }
+    case 'managed': {
+      sponsorAddressHash = deriveSponsorAddressHashForManagedFeed(dapiNameOrDataFeedId);
+      break;
+    }
+    case 'fixed': {
+      sponsorAddressHash = walletDerivationScheme.sponsorAddress!;
```
Siegrift: This is very awkward to read. I think the concept of `sponsorAddressHash` doesn't make sense anymore. It seems better to define `sponsorAddress` and do something like this:
```
    case 'self-funded': {
      const sponsorAddressHash = deriveSponsorAddressHashForSelfFundedFeed(dapiNameOrDataFeedId, updateParameters);
      sponsorAddress = ethers.getAddress(sponsorAddressHash.slice(0, 42)) as Address;
      break;
    }
    ...
```

Siegrift: Also, the documentation needs to be updated for this change.

test/fixtures/mock-contract.ts
```diff
@@ -4,6 +4,11 @@ import { ethers } from 'ethers';
 import { encodeDapiName } from '../../src/utils';
 import { type DeepPartial, encodeBeaconDetails } from '../utils';
 
+const contractFunction = Object.assign(jest.fn(), {
```
Siegrift: nit: I actually liked using `.send` as well instead of fn call. But I don't mind this either.

Siegrift (changes requested):

src/gas-price/gas-price.ts
```diff
@@ -81,9 +81,27 @@ export const fetchAndStoreGasPrice = async (
...
+    pendingTransactionInfos.sort((a, b) => b.consecutivelyUpdatableCount - a.consecutivelyUpdatableCount),
```
bdrhn9: I think we don't need to sort by `consecutivelyUpdatableCount` because we are just getting minimum by `firstUpdatableTimestamp`.

src/gas-price/gas-price.ts
bdrhn9: Probably this comment is reasonably correct but not consistent with the code. Code calculates the oldest pending transaction info according to `firstUpdatableTimestamp` rather than `consecutivelyUpdatableCount`.

local-test-configuration/scripts/initialize-chain.ts
acenolaza: Not really. I just wanted to be consistent since BaseContract and BaseWallet both inherit from Addresseable. I'll revert to just use `.address` property on Wallets

src/gas-price/gas-price.ts
acenolaza: Ended up removing the `null` or `undefined` filtering altogether in [here](https://github.com/api3dao/airseeker-v2/pull/296/commits/d958d2f5a25ddb97748cd4b93dea517354cb2e27)

src/update-feeds-loops/submit-transactions.test.ts
acenolaza: I just wanted to be consistent with how we call contract methods on all other projects using ethers.js. I'll revert this change because I don't care that much tho

src/update-feeds-loops/submit-transactions.ts
acenolaza: Same as [this](https://github.com/api3dao/airseeker-v2/pull/296#discussion_r1594510718)

src/update-feeds-loops/submit-transactions.ts
acenolaza: It seemed obvious to me that wallet parameter must be connected to a provider if is going to be used within `submitUpdate()` to send a tx

Siegrift: True.

local-test-configuration/scripts/initialize-chain.ts
Siegrift: Btw. these "TODO" were for me so I don't forget (but I forgot :D). I wanted to say to mirror the implementation in Airseeker.

local-test-configuration/scripts/initialize-chain.ts
Siegrift: Btw. these "TODO" were for me so I don't forget (but I forgot :D). I wanted to say to mirror the implementation in Airseeker.

local-test-configuration/scripts/initialize-chain.ts
Siegrift: I was OK with `await` ing here, but was curious if there is some particular reason for it.

src/update-feeds-loops/submit-transactions.ts
Siegrift: Thinking about this again, maybe this comment is not really worth it. Feel free to just resolve the comment if you think it's not wroth it.

local-test-configuration/scripts/initialize-chain.ts
acenolaza: I extracted it to a module function but then realized you probably  meant to use `utils.getDerivedSponsorWallet()` 🤭

local-test-configuration/scripts/initialize-chain.ts
acenolaza: Sorry, that answer was actually about using the `async getAddress()` over the `address` property. The deploy() function accepts either `string` or `Promise<string>` like you mentioned so I just reverted the change.

src/config/schema.ts
acenolaza: Thanks for providing the actual code. I added .strict() on each object since we were already using .strict() before this change (and it is nice because it will not allow `sponsorAddress` to be present when `type` is `managed` or `self-funded` for example)

src/update-feeds-loops/submit-transactions.ts
acenolaza: It is. I'm just filling the array with the result of this single call to not mess with the [`successCount`](https://github.com/api3dao/airseeker-v2/blob/single-sponsor-wallet/src/update-feeds-loops/update-feeds-loops.ts#L303) since it is based on data feed updates.

src/update-feeds-loops/submit-transactions.ts
acenolaza: Good point and I have now just realized that I need to also deal with `state.derivedSponsorWallets` which are being keyed by `dapiNameOrDataFeedId` 🤔 I'm even wondering if this map works as expected for self-funded where we might have more than one PK per `dataFeedId`

acenolaza: I think using the `sponsorAddress` as key might be the way to go for caching `privateKeys`...

local-test-configuration/scripts/initialize-chain.ts
acenolaza: Ended up removing this section and just calling the derive function in utils

src/update-feeds-loops/submit-transactions.ts
acenolaza: Done [here](https://github.com/api3dao/airseeker-v2/pull/296/commits/ad7ac3aa2c0a0254572bccffba7f9ad2903c9a68)

src/utils.ts
acenolaza: Done [here](https://github.com/api3dao/airseeker-v2/pull/296/commits/ad7ac3aa2c0a0254572bccffba7f9ad2903c9a68) but I won't resolve this conversation until I update the documentation

test/fixtures/mock-contract.ts
acenolaza: https://github.com/api3dao/airseeker-v2/pull/296/commits/40093f5c73e82073f288caa0e860fb2c565aea7b

src/update-feeds-loops/submit-transactions.ts
acenolaza: I'm now thinking if it's even worth calling [submitTransactions](https://github.com/api3dao/airseeker-v2/blob/single-sponsor-wallet/src/update-feeds-loops/update-feeds-loops.ts#L294) when this array is empty 🤔

src/config/schema.ts
```diff
@@ -154,18 +154,29 @@ export const deviationThresholdCoefficientSchema = z
 
 export type DeviationThresholdCoefficient = z.infer<typeof deviationThresholdCoefficientSchema>;
 
-export const walletDerivationTypeSchema = z.union([z.literal('self-funded'), z.literal('managed')]);
+export const walletDerivationTypeSchema = z.union([z.literal('self-funded'), z.literal('managed'), z.literal('fixed')]);
```
Siegrift: nit: This is unused. There may be an argument for using it inside NPM package, but I'd probably just remove it. If you want to keep it, move `WalletDerivationType` type directly below it.

src/gas-price/gas-price.test.ts
```diff
@@ -312,6 +326,51 @@ describe(getRecommendedGasPrice.name, () => {
       sanitizationGasPriceCap: '19400000000',
     });
   });
+
+  it('applies scaling based on oldest non-null pending data feed when multiple data feed IDs are passed', () => {
```
Siegrift, line 330: 👍

src/gas-price/gas-price.ts
```diff
@@ -81,9 +81,20 @@ export const fetchAndStoreGasPrice = async (
...
+  providerName: string,
+  sponsorWalletAddress: Address,
+  dataFeedIds: Hex[]
+) => {
   const state = getState();
-  const pendingTransactionInfo = state.pendingTransactionsInfo[chainId]![providerName]![sponsorWalletAddress];
+  const oldestPendingTransactionInfo = minBy(
+    dataFeedIds
+      .map((dataFeedId) => state.pendingTransactionsInfo[chainId]?.[providerName]?.[sponsorWalletAddress]?.[dataFeedId])
+      .sort((a, b) => (b?.consecutivelyUpdatableCount ?? 0) - (a?.consecutivelyUpdatableCount ?? 0)), // Sort by consecutivelyUpdatableCount
```
Siegrift: You should sort by `firstUpdatableTimestamp`. The `consecutivelyUpdatableCount` is only for informational purposes (although due to Airseeker design using that would work as well).

Also, I don't see the value in the comment.

src/state/state.ts
Siegrift: I mean, the comment should explain why we map by both sponsor wallet and data feed ID.

src/types.ts
```diff
@@ -16,3 +18,23 @@ export const signedApiResponseSchema = z.object({
   count: z.number().positive(),
   data: z.record(signedDataSchema),
 });
+
+interface BaseParams {
```
Siegrift: This is too generic name to be in `types.ts`. I think it should be moved to `utils.ts`. This would make a bit cleaner:
```
export type SelfFundedParams = SelfFundedWalletDerivationSchemeType & {
  updateParameters: string;
  dapiNameOrDataFeedId: string;
};

export type ManagedParams = ManagedWalletDerivationSchemeType & {
  dapiNameOrDataFeedId: string;
};

export type FixedParams = FixedWalletDerivationSchemeType;

export type SponsorAddressDerivationParams = FixedParams | ManagedParams | SelfFundedParams;

export const deriveSponsorAddress = (params: SponsorAddressDerivationParams) => {
  let sponsorAddress: Address;
  switch (params.type) {
    case 'self-funded': {
      // Airseeker will derive a sponsor wallet for updating each dAPI name or data feed + update parameters combination.
      const { dapiNameOrDataFeedId, updateParameters } = params;
      sponsorAddress = deriveSponsorAddressForSelfFundedFeed(dapiNameOrDataFeedId, updateParameters);
      break;
    }
    case 'managed': {
      // Here it will derive a single sponsor wallet for updating  each dAPI name or data feed.
      const { dapiNameOrDataFeedId } = params;
      sponsorAddress = deriveSponsorAddressForManagedFeed(dapiNameOrDataFeedId);
      break;
    }
    case 'fixed': {
      // Here it will derive a single sponsor wallet for updating all dAPI names or data feeds.
      sponsorAddress = params.sponsorAddress!;
      break;
    }
  }

  return sponsorAddress;
};
```

but ultimately, this just complicates the code, because the callers will need to make sure they pass the correct parameters each time (making the code verbose). I think the best is to just pass all that is needed to `deriveSponsorAddress` and let it pick&choose the correct parameters for derivation.

src/update-feeds-loops/submit-transactions.ts
Siegrift: I see. The code was a bit more cryptic to me and I misread it.

src/utils.test.ts
```diff
@@ -1,28 +1,25 @@
...
+  deriveSponsorAddressForSelfFundedFeed,
   deriveSponsorWallet,
-  deriveSponsorWalletFromSponsorAddressHash,
+  deriveSponsorWalletFromSponsorAddress,
   encodeDapiName,
 } from './utils';
 
-describe(deriveSponsorWalletFromSponsorAddressHash.name, () => {
-  describe(deriveSponsorAddressHashForManagedFeed.name, () => {
+describe(deriveSponsorWalletFromSponsorAddress.name, () => {
```
Siegrift, line 12: Let's add a test for fixed "sponsor wallet derivation" here as well.

src/update-feeds-loops/submit-transactions.ts
```diff
@@ -91,9 +92,109 @@ export const submitUpdate = async (
...
+        const goSubmitUpdate = await go(async () => {
+          logger.info('Updating data feeds in batch.', {
+            sponsorWalletAddress,
+            gasPrice: gasPrice.toString(),
+            gasLimit: gasLimit.toString(),
+            nonce,
+          });
+          return api3ServerV1
+            .connect(sponsorWallet)
+            .tryMulticall.send(dataFeedUpdateCalldatas, { gasPrice, gasLimit, nonce });
```
Siegrift: Btw. we should only use mutlicall in case there are multiple calldatas. It is possible that only a single beacon will be updated and this should be called directly. Feel free to open separate issue for that.

Siegrift (commented):
FYI: I've edited the spec about the wallet derivation types [here](https://docs.google.com/document/d/1x5QBOGII8IUGjtoNR6PVE_UeqEjRQj2u3Ysa1FQkHf0/edit?pli=1#heading=h.8nz2bd8lo3b9)

src/update-feeds-loops/submit-transactions.ts
acenolaza: Right, good catch. Last night I was trying to make both `submitTransaction()` and `submitBatchTransaction()` call a single `submitUpdate()` function that handles both cases based on [this comment](https://github.com/api3dao/airseeker-v2/pull/296#discussion_r1593670739) so maybe I can try avoid sending a tryMulticall in there as well

src/gas-price/gas-price.ts
acenolaza: So this is a similar comment to the one @bdrhn9 made [here](https://github.com/api3dao/airseeker-v2/pull/296#discussion_r1593936792). I ended up sorting by  `consecutivelyUpdatableCount` because each feed in a batch will have the same [firstUpdatableTimestamp](https://github.com/api3dao/airseeker-v2/blob/single-sponsor-wallet/src/update-feeds-loops/pending-transaction-info.ts#L45) therefore I made the assumption that the data feed that has been retried the most should be oldest. Maybe I should sort by both properties 🤔

src/types.ts
acenolaza: Agree. I went for the simpler option that requires all properties to be defined in the `params` object (added a comment for when properties are empty strings using fixed wallet derivation type)

src/config/schema.ts
acenolaza: Removed all unused

Siegrift (commented):
Can you rebase please? I'll review then.

airseeker_v2_pipeline.drawio
Siegrift, line 1: What changed here? (From the diff it seems like only whitespace).

src/gas-price/gas-price.ts
Siegrift: Nah, just sort by the `firstUpdatableTimestamp`. The `consecutivelyUpdatableCount` is really only for infomational purposes (and maybe later for alerting). But even better, as Bedirhan pointed out - do not sort at all, because you use `minBy` anyway.

src/update-feeds-loops/submit-transactions.ts
Siegrift: Doesn't matter to me.

Siegrift (approved):
👍 Implementation looks good to me, but wait for @bdrhn9 's approve as well as this is a quite a significant change.

I think we already talked about this internally, but this should be tested on testnets before we attempt to roll it out for mainnets.

airseeker_v2_pipeline.drawio
acenolaza: Very minor change on line [197](https://github.com/api3dao/airseeker-v2/pull/296/files#diff-956c21e445d3fa875055c31a47aeea1221a8c86d1276b2a7079a41e01cfe495fR197)

Siegrift: I see. Thanks for looking into this as well. Btw. check the spec whether it's up to date and whether it makes sense.

src/state/state.ts
```diff
@@ -21,11 +21,19 @@ export interface PendingTransactionInfo {
 export interface State {
   config: Config;
   gasPrices: Record<ChainId, Record<string /* Provider name */, GasPriceInfo[]>>;
+  // This is used to keep track of the first time Airseeker detects that a data feed requires an update.
+  // The object with the update information is grouped by sponsor wallet and data feed ID. This is because Airseeker uses different schemes to derive sponsor wallets:
+  // `managed`: The sponsor wallet is derived from the dAPI name resulting in a single sponsor wallet per data feed.
+  // `self-funded`: The sponsor wallet is derived from the dAPI name and update parameters resulting in multiple sponsor wallets per data feed.
```
bdrhn9:
```suggestion
  // `self-funded`: The sponsor wallet is derived from the dAPI name (or data feed ID) and update parameters resulting in multiple sponsor wallets per data feed.
```

src/update-feeds-loops/pending-transaction-info.ts
```diff
@@ -17,10 +17,17 @@ export const setPendingTransactionInfo = (
   chainId: string,
   providerName: string,
   sponsorWalletAddress: Address,
+  dataFeedId: Hex,
   pendingTransactionInfo: PendingTransactionInfo | null
 ) => {
   updateState((draft) => {
-    draft.pendingTransactionsInfo[chainId]![providerName]![sponsorWalletAddress] = pendingTransactionInfo;
+    draft.pendingTransactionsInfo[chainId]![providerName] = {
```
bdrhn9: Maybe following is more readable but it's just nit:

```typescript
if (!draft.pendingTransactionsInfo[chainId]![providerName]![sponsorWalletAddress]) {
      draft.pendingTransactionsInfo[chainId]![providerName]![sponsorWalletAddress] = {};
    }
draft.pendingTransactionsInfo[chainId]![providerName]![sponsorWalletAddress]![dataFeedId] = pendingTransactionInfo;
```

bdrhn9 (approved):
Changes looks clean and good to me 💪🏼

src/update-feeds-loops/submit-transactions.ts
```diff
@@ -199,37 +244,38 @@ export const submitTransactions = async (
...
   blockNumber: number
-) =>
-  Promise.all(
+) => {
+  const {
+    config: { walletDerivationScheme },
+  } = getState();
+  if (walletDerivationScheme.type === 'fixed') {
+    return updatableDataFeeds.length > 0
+      ? Array.from({ length: updatableDataFeeds.length }).fill(
```
bdrhn9: I only wasn't satisfied with this part looks too hacky for me. If one doesn't know `.fill` implementation like me can think that `.fill` calls its input for length of array that is going to be filled. Additionally, with this addition, typescript couldn't resolve the output type of main function `submitTransactions` and just says `unknown`. 

I checked that we are only using output of this function to calculate success count [here](https://github.com/api3dao/airseeker-v2/blob/d17db3c82c37c612a5971276394fb35bf6502e02/src/update-feeds-loops/update-feeds-loops.ts#L313). So my alternative solution would be implementing this function similar to following. Wdyt?

```typescript
export const submitTransactions = async (
  chainId: string,
  providerName: string,
  provider: ethers.JsonRpcProvider,
  api3ServerV1: Api3ServerV1,
  updatableDataFeeds: UpdatableDataFeed[],
  blockNumber: number
) => {
  if (updatableDataFeeds.length === 0) return 0;
  const {
    config: { walletDerivationScheme },
  } = getState();

  if (walletDerivationScheme.type === 'fixed') {
    const result = await submitBatchTransaction(
      chainId,
      providerName,
      provider,
      api3ServerV1,
      updatableDataFeeds,
      blockNumber
    );
    return result ? updatableDataFeeds.length : 0;
  }

  const result = await Promise.all(
    updatableDataFeeds.map(async (dataFeed) =>
      submitTransaction(chainId, providerName, provider, api3ServerV1, dataFeed, blockNumber)
    )
  );
  return result.filter(Boolean).length;
};
```

acenolaza: Looks really neat 🔥 Thanks for the taking the time to actually write the code. I'll apply that change and fix the tests 🙏🏻

acenolaza:
Thank you both for all the feedback! 🚀

acenolaza:
Thank you both for all the feedback! 🚀 

Should I just merge now? How should we test these changes? Do you guys deploy to a testnet to verify everything is still working as expected?

bdrhn9:
> Thank you both for all the feedback! 🚀
> 
> Should I just merge now? How should we test these changes? Do you guys deploy to a testnet to verify everything is still working as expected?

I think changes looks good and mergeable. I will test these changes by deploying on one testnet. Maybe I'll remove actual deployment on that specific testnet to see whether our fallback airseeker gain control over.

PR #301, opened 2024-05-08
Do not validate Signed API responses from config

### LINKED ISSUE #300: Do not validate responses from config specified Signed API
See: https://github.com/api3dao/airseeker-v2/issues/282#issuecomment-2097962172

Decided to open a new issue because https://github.com/api3dao/airseeker-v2/pull/291 is nearly finished and this one will have a bit more changes.

Siegrift:
Closes https://github.com/api3dao/airseeker-v2/issues/300

### DIFF
--- a/reset.d.ts
+++ b/reset.d.ts
@@ -1,0 +1,2 @@
+// Do not add any other lines of code to this file!
+import '@total-typescript/ts-reset';
--- a/src/data-fetcher-loop/data-fetcher-loop.ts
+++ b/src/data-fetcher-loop/data-fetcher-loop.ts
@@ -60,7 +60,8 @@ export const runDataFetcher = async () => {
     const state = getState();
     const {
       config: { signedDataFetchInterval },
-      signedApiUrls,
+      signedApiUrlsFromConfig,
+      signedApiUrlsFromContract,
       activeDataFeedBeaconIds,
     } = state;
     const signedDataFetchIntervalMs = signedDataFetchInterval * 1000;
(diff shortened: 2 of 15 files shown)


src/data-fetcher-loop/data-fetcher-loop.ts
```diff
@@ -73,10 +74,17 @@ export const runDataFetcher = async () => {
         .flat(2)
     );
 
+    // Compute the set of URLs coming from the config. These are trusted and don't need to be verified.
```
Siegrift, line 77: The question is what to do if there is a same URL in both contract and config. It makes sense to treat it as trusted, so that is what the code does as well.

reset.d.ts
```diff
@@ -0,0 +1,2 @@
+// Do not add any other lines of code to this file!
+import '@total-typescript/ts-reset';
```
Siegrift, line 2: This allows for a nicer `array.filter(Boolean)` inference and a few other TS defaults tweaks.

src/data-fetcher-loop/data-fetcher-loop.ts
```diff
@@ -73,10 +74,17 @@ export const runDataFetcher = async () => {
...
+        .map((urlsPerProvider) => Object.values(urlsPerProvider))
+        .flat(2)
+    );
+
     // Better to log the non-decomposed object to see which URL comes from which chain-provider group.
-    logger.debug('Signed API URLs.', { signedApiUrls });
+    logger.debug('Signed API URLs.', { signedApiUrlsFromConfig, signedApiUrlsFromContract });
     const urls = uniq(
-      Object.values(signedApiUrls)
+      [...Object.values(signedApiUrlsFromConfig), ...Object.values(signedApiUrlsFromContract)]
```
Siegrift, line 87: This is the key part of the PR - we now separate the config and contract specific URLs and merge them here.

reset.d.ts
bdrhn9: Wow it's nice to have 👍🏼

src/data-fetcher-loop/data-fetcher-loop.ts
bdrhn9: I agree with your decision.

bdrhn9 (approved):
I couldn't find anything to add, it just looks good to me 👍🏼

PR #313, opened 2024-05-21
Fix entry point for package.json

### LINKED ISSUE #312: Fix incorrect package entry point
I [encountered](https://github.com/api3dao/data-feeds/pull/191#issuecomment-2122392680) an issue while trying to import packages after installing @api3/airseeker from [npm](https://www.npmjs.com/package/@api3/airseeker). As I discovered the entry point [here](https://github.com/api3dao/airseeker-v2/blob/31bac37ff87c568e925d3f436fcf4315ae444984/package.json#L13) should be `./dist/src/npm-exports.js` rather than `./dist/npm-exports.js`.

bdrhn9:
Closes #312

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "url": "https://github.com/api3dao/airseeker-v2"
   },
   "scripts": {
-    "build": "rm -rf dist && tsc --project tsconfig.build.json",
+    "build": "tsc --project tsconfig.build.json",
     "clean": "rm -rf coverage dist artifacts cache",
     "create-release:docker": "ts-node scripts/create-docker-release.ts",
     "create-release:npm": "ts-node scripts/create-npm-release.ts",
--- a/package.json
+++ b/package.json
@@ -10,14 +10,14 @@
     "dist",
     "src"
   ],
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -10,14 +10,14 @@
...
-  "main": "./dist/npm-exports.js",
+  "main": "./dist/src/npm-exports.js",
   "sideEffects": false,
   "repository": {
     "type": "git",
     "url": "https://github.com/api3dao/airseeker-v2"
   },
   "scripts": {
-    "build": "tsc --project tsconfig.build.json",
+    "build": "rm -rf dist && tsc --project tsconfig.build.json",
```
Siegrift: We should either run `pnpm run clean` or not run this at all (and only run the `pnpm run clean` in the release instructions).

Siegrift (approved):
Yeah, our `dist` folder contains `src` folder so this change 👍 LGTM - thanks.

package.json
bdrhn9: Okay I reverted change on this script and merging now. Thanks for the review 💪🏼

PR #315, opened 2024-05-23
Prevent gas price scaling if data feed was updated since the first update tx

### LINKED ISSUE #295: Prevent gas price scaling if Airseeker update went through
When the market is volatile, there is a possibility that a feed becomes updatable even though it was recently updated by Airseeker. Currently, it's possible for Airseeker to confuse this case with a "pending transaction", causing Airseeker to scale the gas price unnecessarily.

See [slack thread](https://api3workspace.slack.com/archives/C05TQPT7PNJ/p1713458319193379) for details.

Siegrift:
Closes https://github.com/api3dao/airseeker-v2/issues/295

## Rationale

The idea is to persist the on-chain timestamp of the data feed at the time of the transaction. When Airseeker sees that the feed requires consecutive update, we check whether the timestamp has changed. If the timestamp has changed, we know the feed was updated - either by the previous transaction or other Airseeker update, but the update source does not really matter. In that case, we reset the pending transaction info and consider the next update as original.

As a note, we track pending transaction info per sponsor per data feed to support different wallet derivation schemes, but [updates are data feed specific](https://api3workspace.slack.com/archives/C05TQPT7PNJ/p1713948371457569?thread_ts=1713458319.193379&cid=C05TQPT7PNJ). So if there are pending transactions for the same data feed from two sponsor wallets, their pending transaction info should be reset if the feed was updated. The current scheme handles that implicitly.

### DIFF
--- a/src/gas-price/gas-price.test.ts
+++ b/src/gas-price/gas-price.test.ts
@@ -228,6 +228,7 @@ describe(getRecommendedGasPrice.name, () => {
           [dataFeedId]: {
             consecutivelyUpdatableCount: 2,
             firstUpdatableTimestamp: timestampMock - 60, // The feed requires update for 1 minute.
+            onChainTimestamp: BigInt(timestampMock - 65),
           },
         },
       };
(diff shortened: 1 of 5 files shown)


bdrhn9 (approved):
LGTM 💪🏼

PR #321, opened 2024-05-27
Fetch gas price using low-level RPC method

### LINKED ISSUE #305: Fetch gas price natively
See: https://api3workspace.slack.com/archives/C06R2JCFSS1/p1715585624116669?thread_ts=1715333508.842309&cid=C06R2JCFSS1

bdrhn9:
Closes #305

### DIFF
--- a/src/gas-price/gas-price.test.ts
+++ b/src/gas-price/gas-price.test.ts
@@ -1,5 +1,5 @@
 import type { Hex } from '@api3/commons';
-import { ethers } from 'ethers';
+import { ethers, toBeHex } from 'ethers';
 import { range } from 'lodash';
 
 import { generateTestConfig, initializeState } from '../../test/fixtures/mock-config';
@@ -7,7 +7,16 @@ import { logger } from '../logger';
 import { getState, updateState } from '../state';
 import { initializePendingTransactionsInfo } from '../update-feeds-loops/pending-transaction-info';
 
-import * as gasPriceModule from './gas-price';
+import {
+  getRecommendedGasPrice,
+  saveGasPrice,
+  purgeOldGasPrices,
+  initializeGasState,
+  calculateScalingMultiplier,
+  getPercentile,
+  fetchAndStoreGasPrice,
+  getGasPrice,
+} from './gas-price';
 
 const chainId = '31337';
 const providerName = 'localhost';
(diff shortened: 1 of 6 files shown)


bdrhn9:
Alternatively, it can be implemented by extending the JsonRpcProvider class like following:
```typescript
// eslint-disable-next-line functional/no-classes
export class LegacyGasPriceRpcProvider extends ethers.JsonRpcProvider {
  public async getGasPrice(): Promise<bigint> {
    const value = await this.send('eth_gasPrice', []);

    return getBigInt(value);
  }
}
```

I didn't prefer this approach because it would necessitate modifying nearly every file that references `ethers.JsonRpcProvider`, both as a class and as an interface.

src/update-feeds-loops/update-feeds-loops.test.ts
```diff
@@ -443,6 +444,7 @@ describe(updateFeedsLoopsModule.processBatch.name, () => {
     jest.spyOn(logger, 'info');
 
     // Skip actions other than generating signed api urls.
+    jest.spyOn(gasPriceModule, 'fetchAndStoreGasPrice').mockImplementation();
```
Siegrift, line 512: Mhh, why do these need to be mocked now?

src/gas-price/gas-price.test.ts
```diff
@@ -373,12 +365,22 @@ describe(getRecommendedGasPrice.name, () => {
...
+    expect(provider.send).toHaveBeenCalledWith('eth_gasPrice', []);
+    expect(gasPrice).toBe(BigInt('0x1a13b8600'));
+  });
+});
+
+describe(gasPriceModule.fetchAndStoreGasPrice.name, () => {
   it('fetches and stores the gas price from RPC provider', async () => {
     jest.spyOn(Date, 'now').mockReturnValue(dateNowMock);
-    jest.spyOn(provider, 'getFeeData').mockResolvedValueOnce({ gasPrice: ethers.parseUnits('10', 'gwei') } as any);
+    jest.spyOn(gasPriceModule, 'getGasPrice').mockResolvedValueOnce(ethers.parseUnits('10', 'gwei'));
```
Siegrift: I believe you can spy on the provider instead and this way you avoid the need to import the entirety of gas price module. (Similarly in the test below).

src/gas-price/gas-price.test.ts
```diff
@@ -373,12 +365,22 @@ describe(getRecommendedGasPrice.name, () => {
   });
 });
 
-describe(fetchAndStoreGasPrice.name, () => {
+describe(gasPriceModule.getGasPrice.name, () => {
+  it('returns gas price as bigint when internal rpc call returns a valid hex string', async () => {
```
Siegrift, line 378: Can you add one more test when the provider fails? I checked [Ethereum RPC spec](https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_gasprice) (and some other sources) and I think we can rely that when the API returns something it will be a hex string...

Siegrift (commented):
👍 LGTM

I agree with your implementation choice. This is simple enough.

src/update-feeds-loops/update-feeds-loops.test.ts
bdrhn9: Yes it looks redundant but I wanted to add them because `processBatch` includes `fetchAndStoreGasPrice`. Should I remove them?

src/gas-price/gas-price.test.ts
bdrhn9: Yep, that is what I did first. However, with the current approach, it is abstracted from provider and depends on `getGasPrice` which we already tested. I'm not good at tests 😄 I'll do what you suggest at the end of the conversation.

src/update-feeds-loops/update-feeds-loops.test.ts
Siegrift: Nah, leave them. I was just surprised why the test was passing, but I guess either Ethers made a noop call or silently did nothing 🤷

src/gas-price/gas-price.test.ts
Siegrift: I see the point, but I think the ugliness it brings is not worth it. We can have a similarly readable test with:
```
jest.spyOn(provider, 'send').mockResolvedValueOnce(ethers.toBeHex(ethers.parseUnits('10', 'gwei')));
```

src/update-feeds-loops/update-feeds-loops.test.ts
bdrhn9: Test was passing probably because `fetchAndStoreGasPrice` handles error by returning null when it fails. However, this spy on `fetchAndStoreGasPrice` is good to have because I encountered [this](https://github.com/ethers-io/ethers.js/blob/ad5f1c5fc7b73cfb20d9012b669e9dcb9122d244/src.ts/providers/provider-jsonrpc.ts#L767) error when I am playing with tests even though I couldn't produce it now.

Siegrift (approved):
👍 LGTM

PR #322, opened 2024-05-28
Use block number from the RPC call

### LINKED ISSUE #314: Use smaller block number - from the contract and from the RPC
See [thread](https://api3workspace.slack.com/archives/C05TQPT7PNJ/p1715963064307619).

There are some chains, for which the contract block number is slightly higher than the block number returned by the RPC. The problem is that when this happens, we fail to get nonce (and delay the update considerably).

We've agreed on the call to use the smaller of the block number from the contract and by the RPC. The RPC call is made after the contract call. There is a possibility of a race condition, but we will see if this turns out to be a problem in practice.

bdrhn9:
Closes #314

### DIFF
--- a/src/update-feeds-loops/contracts.ts
+++ b/src/update-feeds-loops/contracts.ts
@@ -50,8 +50,6 @@ export const verifyMulticallResponse = (
 
 export const decodeActiveDataFeedCountResponse = Number;
 
-export const decodeGetBlockNumberResponse = Number;
-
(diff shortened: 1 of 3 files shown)


Siegrift:
@bdrhn9 Next time base the PR on the other branch, e.g.

image

(That makes it a bit easier to review)

Siegrift (approved):
👍 LGTM

PR #334, opened 2024-06-20
Add heartbeat interval modifier

### LINKED ISSUE #330: Implement `heartbeatIntervalModifier`
See https://github.com/api3dao/airseeker-v2/issues/80 and https://api3workspace.slack.com/archives/C05TQPT7PNJ/p1696499772685759

The value can be positive or negative. The resulting heartbeat interval cannot be negative.

Not having it means the fallback Airseeker will frontrun the primary one 50% of the time for heartbeats.

bdrhn9:
Closes #330.

### DIFF
--- a/config/configuration.md
+++ b/config/configuration.md
@@ -203,7 +203,9 @@ Defaults to `1`.
 
 #### `heartbeatIntervalModifier` _(optional)_
 
-The global modifier applied to all heartbeat checks. The value is specified in seconds and can be positive or negative. The resulting interval is calculated by adding this modifier as an offset and clamped to a minimum of 0, ensuring it never goes negative. Used to differentiate alternate deployments. For example:
+The global modifier applied to all heartbeat checks. The value is specified in seconds and can be positive or negative.
+The resulting interval is calculated by adding this modifier as an offset and clamped to a minimum of 0, ensuring it
+never goes negative. Used to differentiate alternate deployments. For example:
 
 ```jsonc
 "heartbeatIntervalModifier": -120,
--- a/config/configuration.md
+++ b/config/configuration.md
@@ -203,8 +203,7 @@ Defaults to `1`.
 
 #### `heartbeatIntervalModifier` _(optional)_
 
-The global modifier applied to all heartbeat checks. Used to differentiate alternate deployments. In seconds. For
-example:
+The global modifier applied to all heartbeat checks. The value is specified in seconds and can be positive or negative. The resulting interval is calculated by adding this modifier as an offset and clamped to a minimum of 0, ensuring it never goes negative. Used to differentiate alternate deployments. For example:
 
 ```jsonc
 "heartbeatIntervalModifier": -120,
--- a/config/configuration.md
+++ b/config/configuration.md
@@ -201,6 +201,17 @@ The global coefficient applied to all deviation checks. Used to differentiate al
 
 Defaults to `1`.
 
+#### `heartbeatIntervalModifier` _(optional)_
+
+The global modifier applied to all heartbeat checks. Used to differentiate alternate deployments. In seconds. For
+example:
+
+```jsonc
+"heartbeatIntervalModifier": -120,
+```
+
+Defaults to `0`.
+
 #### `dataFeedUpdateInterval`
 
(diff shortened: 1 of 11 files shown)


src/deviation-check/deviation-check.ts
```diff
@@ -60,13 +62,13 @@ export const checkUpdateCondition = (
 ): boolean => {
   if (onChainTimestamp === 0n && offChainTimestamp > 0) return true;
   if (
-    deviationThreshold &&
```
bdrhn9, line 63: Actually didn't get the first part of the condition check. Shouldn't we already expect it to be bigint as function input parameters declares. I utilized `isNil` otherwise value `0n` for `deviationThreshold` or `heartbeatInterval` cause to not satisfy if statements in advance. But I think `isNil` can also be removed.

src/update-feeds-loops/get-updatable-feeds.ts
```diff
@@ -83,6 +84,18 @@ export const getUpdatableFeeds = (
...
+        if (modifiedHeartbeatInterval < 0n) {
+          logger.warn('Resulting heartbeat interval is negative. Setting it to the original value.', {
+            dapiName: decodedDapiName,
+            dataFeedId,
+            heartbeatInterval,
+            heartbeatIntervalModifier,
+          });
+        }
+        const adjustedHeartbeatInterval =
+          modifiedHeartbeatInterval < 0n ? heartbeatInterval : modifiedHeartbeatInterval;
```
bdrhn9: When the resulting heartbeat is negative, I opted to set it to original value. But it can be preferred to set it to 0.

config/configuration.md
```diff
@@ -201,6 +201,17 @@ The global coefficient applied to all deviation checks. Used to differentiate al
 
 Defaults to `1`.
 
+#### `heartbeatIntervalModifier` _(optional)_
+
+The global modifier applied to all heartbeat checks. Used to differentiate alternate deployments. In seconds. For
```
Siegrift: I think it would be nice to document that it can be positive/negative and also what exactly it does.

src/deviation-check/deviation-check.ts
Siegrift: I believe I wanted to prevent cases when the `deviationThreshold` and `heartbeatInterval` were uninitialised, but not sure if this is even possible.

But I don't see an use case for either of the values to be set to 0.

src/update-feeds-loops/get-updatable-feeds.ts
Siegrift: I'd say by doing specifying negative modifier, you essentially want to decrease the heartbeat interval, so by that logic setting it to 0 means we set it to the lowest reasonable value. That said, it doesn't make sense to me for it to be 0.

src/deviation-check/deviation-check.ts
bdrhn9: Maybe there will be a feed that expected to be updated in each cycle by setting heartbeat interval or deviation threshold to 0.

So I'm keeping my changes if the reasoning is incase of uninitialization.

src/update-feeds-loops/get-updatable-feeds.ts
bdrhn9: Hmm. Is there better alternative you can suggest?

P.S.: Burak also suggested setting it to 0 if the resulting is negative. For the sake of democracy, I'll set it to 0 😄

src/deviation-check/deviation-check.ts
Siegrift: If that is legit requirement, then those lines should imo be removed.

src/update-feeds-loops/get-updatable-feeds.ts
Siegrift: My argument is basically the same as [here](https://github.com/api3dao/airseeker-v2/pull/334#discussion_r1648744658) that I didn't see the use case for that. I am fine with implementation allowing that.

Siegrift (approved):
👍 LGTM (but I think we should remove those checks as they are extraneous)

PR #336, opened 2024-06-23
Export package version

### LINKED ISSUE #335: Export package version
When creating the Airseeker configuration [here](https://github.com/api3dao/data-feeds/blob/a7bf640e47a3ae13c3016be143d25519e3d19728/packages/airseeker-admin/scripts/generate-airseeker-config.ts#L66), the configuration version needs to be filled in manually with each @api3/airseeker NPM package update. Instead, it would be very helpful to export the package version as property, similar to how [lodash](https://lodash.com/docs/4.17.15#VERSION) does it.

bdrhn9:
Closes #335

### DIFF
--- a/src/config/schema.ts
+++ b/src/config/schema.ts
@@ -4,7 +4,10 @@ import { addressSchema } from '@api3/commons';
 import { ethers } from 'ethers';
 import { z } from 'zod';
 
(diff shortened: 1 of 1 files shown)


Siegrift (approved):
👍 LGTM

PR #343, opened 2024-07-02
Async beaconset beacons update

### LINKED ISSUE #329: Update individual Beacons if they deviate too much, independent from the aggregation
Api3ServerV1 expects individual Beacons comprising Beacon sets to be kept up to date independent from the aggregation. For example, imagine three Beacons with the following signed data
```
t=0 100, 100, 100
t=1 100, 50, 100
t=2 100, 100, 100
```
Here, the aggregation of the Beacon sets doesn't change, so Airseeker will not update. However, this causes the data point 50 to remain usable in any following update, which is not good. Instead, Airseeker should update the second Beacon from 100 to 50 (at t=1), back to 100 again (at t=2), which renders the 50 unusable in the following update.

Few points:
- This (allowing asynchronous updates of Beacons) is an intentional tradeoff. We do some redundant Beacon updates in return for resiliency against unavailability.
- It is not unheard of for a single Beacon to spaz out (e.g. JPY/USD that flips between Yen and new Yen conventions). Therefore, it's possible for this to drain the wallet, which means this should only be enabled on the main or the fallback (fixed sponsor wallet) Airseeker configurations. So whether this is enabled should be config-configurable.
- This isn't about preventing numerical inaccuracy (asynchronous updates of Beacons inherently sacrifices that in favor of better availability). Therefore, it can be a lot more relaxed compared to the deviation threshold of the aggregation. In other words, if we're running a dAPI at 1%, we may want to update individual Beacons when their deviation is more than 5% (compared to their on-chain value). This means that this should be config-configurable (probably as a factor of the dAPI deviation threshold).

acenolaza:
Closes https://github.com/api3dao/airseeker-v2/issues/329

### What does this change

This PR tries to enable updating individual beacons of a beacon set data feed that does not require an update. 
1. Most relevant changes are in `getUpdatableFeeds()`. This feature can be enabled by setting a value in config file. This value represents the deviation of individual beacons. A new boolean is returned in each `UpdatableDataFeed` object indicating if `Api3ServerV1.updateBeaconSetWithBeacons()` calldata needs to be added to multicall.
2. For this type of updates I've made an "executive decision" of not setting anything since this is really an edge case and I don't see the value of retrying with higher gas price, etc. Airseeker will print a warning tho when this happens.
3. Used that flag from 1 in `createUpdateFeedCalldatas()` to skip updating the beacon set


config/configuration.md
```diff
@@ -201,6 +201,19 @@ The global coefficient applied to all deviation checks. Used to differentiate al
 
 Defaults to `1`.
 
+#### `asyncBeaconUpdatedeviationThresholdFactor` _(optional)_
```
Siegrift:
```suggestion
#### `asyncBeaconUpdateDeviationThreshold` _(optional)_
```

Siegrift: or `asyncBeaconUpdateDeviationThresholdCoefficient` to be more consistent with deviationThresholdCoefficient

config/configuration.md
```diff
@@ -201,6 +201,19 @@ The global coefficient applied to all deviation checks. Used to differentiate al
...
+
+The global deviation factor applied to async beacon updates deviation checks. When configured, this value will be used
```
Siegrift:
```suggestion
The global deviation factor applied to individual beacon updates deviation checks. When configured, this value will be used
```

config/configuration.md
Siegrift: or ultimately `individualBeaconUpdateDeviationInPercentage` because async does not refer to anything meaningul imo. Also, it's more intuitive to have this in percentage imo.

I think the `deviationThresholdCoefficient` should apply to this as well.

src/update-feeds-loops/get-updatable-feeds.ts
```diff
@@ -19,122 +19,152 @@ export interface UpdatableBeacon {
...
+      logger.warn('Resulting heartbeat interval is negative. Setting it to 0.', {
+        dapiName: decodedDapiName,
+        dataFeedId,
+        heartbeatInterval,
+        heartbeatIntervalModifier,
+      });
+    }
+    const adjustedHeartbeatInterval = modifiedHeartbeatInterval < 0n ? 0n : modifiedHeartbeatInterval;
+
+    // Filter out data feeds that cannot be updated.
```
Siegrift: This comment does not make sense.

src/update-feeds-loops/get-updatable-feeds.ts
```diff
@@ -19,122 +19,152 @@ export interface UpdatableBeacon {
...
+    const isBeaconSet = beaconValues.length > 1;
+
+    // We need to make sure that the update transaction actually changes the on-chain value. There are two cases:
```
Siegrift: Can you wrap the comments according to the line limit. I use this extension https://marketplace.visualstudio.com/items?itemName=stkb.rewrap that does this semi-automatically.

src/update-feeds-loops/get-updatable-feeds.ts
```diff
@@ -19,122 +19,152 @@ export interface UpdatableBeacon {
...
+            newDataFeedValue,
+            newDataFeedTimestamp,
+            adjustedHeartbeatInterval,
+            adjustedDeviationThresholdCoefficient,
+            deviationReference
+          )
+      );
+
+    if (dataFeedNeedsUpdate) {
+      acc.push({
```
Siegrift: You should never push to accumulator, but instead append. That said, this reduce is very complex so I'd suggest writing a for loop instead (and defining the accumulator above the loop and pushing to it).

This also enables you to write early return/continue which reads better.

src/update-feeds-loops/get-updatable-feeds.ts
```diff
@@ -19,122 +19,152 @@ export interface UpdatableBeacon {
...
           .map(({ beaconId, signedData }) => ({
             beaconId,
             signedData: signedData!,
           })),
-      }))
-  );
+        shouldUpdateBeaconSet: isBeaconSet,
+      });
+    } else if (isBeaconSet && asyncBeaconUpdatedeviationThresholdFactor) {
+      // 2.5. There is a special case when data feed is a beacon set that do not need to be updated but some of its beacon constituents do.
```
Siegrift: I consider the "2.5. " a bit misleading and would remove it and keep just the comment.

src/update-feeds-loops/get-updatable-feeds.ts
```diff
@@ -19,122 +19,152 @@ export interface UpdatableBeacon {
...
+          ({ isUpdatable, offChainValue, onChainValue }) =>
+            isUpdatable &&
+            offChainValue &&
+            checkUpdateCondition(
+              onChainValue.value,
+              onChainValue.timestamp,
+              offChainValue.value,
+              offChainValue.timestamp,
+              adjustedHeartbeatInterval,
+              adjustedDeviationThresholdCoefficient * BigInt(asyncBeaconUpdatedeviationThresholdFactor),
```
Siegrift: I think this is too complex. I think it's better to make it a percentage and apply the adjustment factor like I mentioned in the other comment.

src/update-feeds-loops/get-updatable-feeds.ts
```diff
@@ -19,122 +19,152 @@ export interface UpdatableBeacon {
...
           .filter(({ isUpdatable }) => isUpdatable)
           .map(({ beaconId, signedData }) => ({
             beaconId,
             signedData: signedData!,
           })),
-      }))
-  );
+        shouldUpdateBeaconSet: isBeaconSet,
+      });
+    } else if (isBeaconSet && asyncBeaconUpdatedeviationThresholdFactor) {
```
Siegrift: You don't need to check `isBeaconSet` here afaik.

config/configuration.md
Siegrift: I'd maybe suggest making it required, but allow `null` to make it explicit when this is turned off.

config/configuration.md
```diff
@@ -201,6 +201,19 @@ The global coefficient applied to all deviation checks. Used to differentiate al
...
+on beacons sets that don't require an update but some of its beacon constituents do. So if a beacon set has a 1%
+deviation threshold, Airseeker will only update its beacons when deviation of each beacon exeeds (1 \*
+asyncBeaconUpdatedeviationThresholdFactor)%. For example:
+
+```jsonc
+"asyncBeaconUpdatedeviationThresholdFactor": 5,
+```
+
+Defaults to `1`.
```
Siegrift: You're lying 😄 .

config/configuration.md
```diff
@@ -201,6 +201,19 @@ The global coefficient applied to all deviation checks. Used to differentiate al
 
 Defaults to `1`.
 
+#### `asyncBeaconUpdatedeviationThresholdFactor` _(optional)_
+
+The global deviation factor applied to async beacon updates deviation checks. When configured, this value will be used
+on beacons sets that don't require an update but some of its beacon constituents do. So if a beacon set has a 1%
+deviation threshold, Airseeker will only update its beacons when deviation of each beacon exeeds (1 \*
+asyncBeaconUpdatedeviationThresholdFactor)%. For example:
```
Siegrift: The text should also mention:
> - This (allowing asynchronous updates of Beacons) is an intentional tradeoff. We do some redundant Beacon updates in return for resiliency against unavailability.
> - It is not unheard of for a single Beacon to spaz out (e.g. JPY/USD that flips between Yen and new Yen conventions). Therefore, it's possible for this to drain the wallet, which means this should only be enabled on the main or the fallback (fixed sponsor wallet) Airseeker configurations. So whether this is enabled should be config-configurable.

(in a shortened way though).

src/update-feeds-loops/pending-transaction-info.ts
```diff
@@ -40,22 +40,29 @@ export const updatePendingTransactionsInfo = (
...
+    const updatableFeed = feedsToUpdate.find(
       (updatableFeed) =>
         updatableFeed.dataFeedInfo.dapiName === dapiName && updatableFeed.dataFeedInfo.dataFeedId === dataFeedId
     );
 
+    if (updatableFeed && !updatableFeed.shouldUpdateBeaconSet && beaconsWithData.length > 1) {
+      // Feed is a beacon set and it will only update its beacons. Therefore no need to set nor clear pending transaction info.
+      logger.warn(
+        'Data feed is a beacon set that does not require an update but some of its beacons do. Skipping pending transaction info update.',
+        { dapiName: decodedDapiName, dataFeedId }
```
Siegrift, line 53: nit: I think dapiName and dataFeedId may already be logged because of the log context propagation. Not sure about this though.

src/update-feeds-loops/pending-transaction-info.ts
```diff
@@ -40,22 +40,29 @@ export const updatePendingTransactionsInfo = (
   } = getState();
 
   const currentTimestamp = Math.floor(Date.now() / 1000);
-  for (const feed of batch) {
-    const { dapiName, dataFeedId, decodedDapiName, updateParameters, dataFeedTimestamp } = feed;
-
-    const isFeedUpdatable = feedsToUpdate.some(
+  for (const { beaconsWithData, dapiName, dataFeedId, decodedDapiName, updateParameters, dataFeedTimestamp } of batch) {
+    const updatableFeed = feedsToUpdate.find(
```
Siegrift, line 44: nit: can we preserve the old name?

src/update-feeds-loops/submit-transactions.ts
```diff
@@ -31,8 +31,8 @@ export const createUpdateFeedCalldatas = (api3ServerV1: Api3ServerV1, updatableD
     ])
   );
 
-  // If there are multiple beacons in the data feed it's a beacons set which we need to update as well.
-  return allBeacons.length > 1
+  // If there are multiple beacons in the data feed it's a beacons set which we may need to update as well.
+  return allBeacons.length > 1 && shouldUpdateBeaconSet
```
Siegrift: Maybe we can rely only on `shouldUpdateBeaconSet`. By reading the code the other check looks extranous.

Siegrift (changes requested):
Did a first pass and left some comments. I think this should work though.

bdrhn9:
Can you wait me before merge? I'm planning to review tomorrow.

config/configuration.md
bdrhn9: From the last item of the issue:
> This means that this should be config-configurable (probably as a factor of the dAPI deviation threshold).

So I think `asyncBeaconUpdateDeviationThresholdCoefficient` or `individualBeaconUpdateDeviationThresholdCoefficient` may be better choice. That being said, `individualBeaconUpdate...` sounds great, however it doesn't show difference between data feed w/ one beacon and async beacon update.

src/update-feeds-loops/get-updatable-feeds.test.ts
```diff
@@ -572,17 +572,77 @@ describe(getUpdatableFeeds.name, () => {
...
       templateId: '0xe6df5fb43a0b3a65ac1b05e7e50fba03b475fe5b721693d469554278086fd2e4',
     });
-    expect(logger.warn).toHaveBeenCalledWith(`Not using the the signed data because it's older than 24 hours.`, {
+    expect(logger.warn).toHaveBeenCalledWith(`Not using the signed data because it's older than 24 hours.`, {
       airnode: '0xbF3137b0a7574563a23a8fC8badC6537F98197CC',
       templateId: '0x6f0c2b5c6420d1896e67e56539ccbec5e6aafee5c27f6eb8783b9731faa7205d',
     });
   });
+
+  it('returns updatable feeds with beacons that need to be udpated based on asyncBeaconUpdatedeviationThresholdFactor config property', () => {
```
bdrhn9:
```suggestion
  it('returns updatable feeds with beacons that need to be updated based on asyncBeaconUpdatedeviationThresholdFactor config property', () => {
```

src/update-feeds-loops/get-updatable-feeds.test.ts
```diff
@@ -572,17 +572,77 @@ describe(getUpdatableFeeds.name, () => {
...
+        dapiName: encodeDapiName('test'),
+      },
+    ]);
+
+    const checkFeedsResult = getUpdatableFeeds(batch, 1, 0, 5);
+
+    expect(logger.info).toHaveBeenCalledTimes(1);
+    expect(logger.info).toHaveBeenCalledWith(`Deviation exceeded.`);
+    expect(checkFeedsResult).toHaveLength(1);
+    expect(checkFeedsResult[0]!.updatableBeacons).toStrictEqual([
```
bdrhn9, line 638: We can add following check:

`expect(checkFeedsResult[0]!.shouldUpdateBeaconSet).toBe(false);`

src/update-feeds-loops/get-updatable-feeds.ts
```diff
@@ -19,122 +19,152 @@ export interface UpdatableBeacon {
...
+      // 2.5. There is a special case when data feed is a beacon set that do not need to be updated but some of its beacon constituents do.
+      //      In this particular case, airseeker can update only these beacons and skip the beacon set update. This is enabled by setting a
+      //      value in asyncBeaconUpdatedeviationThresholdFactor on the config.
+      const updatableBeacons = aggregatedBeaconsWithData
+        .filter(
+          ({ isUpdatable, offChainValue, onChainValue }) =>
+            isUpdatable &&
+            offChainValue &&
+            checkUpdateCondition(
```
bdrhn9: Maybe it's better to wrap this call with `logger.runWithContext` like above.

src/update-feeds-loops/get-updatable-feeds.ts
```diff
@@ -19,122 +19,152 @@ export interface UpdatableBeacon {
...
+              deviationReference
+            )
+        )
+        .map(({ beaconId, signedData }) => ({
+          beaconId,
+          signedData: signedData!,
+        }));
+      acc.push({
+        dataFeedInfo,
+        updatableBeacons,
```
bdrhn9: Isn't there a chance that `updatableBeacons` is empty array?

bdrhn9 (commented):
With considering @Siegrift comments fulfilled, implementation looks good to me. The only thing that makes me confusing `shouldUpdateBeaconSet`. Because by just looking that one cannot understand it's async beacon update or data feed (w/ one beacon) update. To understand this, one should also look at the length of `.beaconsWithData`. Rather than extending `UpdatableDataFeed` with `shouldUpdateBeaconSet`, we may define `isAsyncBeaconUpdate`.

config/configuration.md
acenolaza: https://github.com/api3dao/airseeker/pull/343/commits/cbc17c7042d0cc1f25c3fe847e588ea65128ae59

src/update-feeds-loops/get-updatable-feeds.ts
acenolaza: https://github.com/api3dao/airseeker/pull/343/commits/4f32637a2bf23bb5e8c20f5aee555cf2655d5b6b

src/update-feeds-loops/get-updatable-feeds.ts
acenolaza: https://github.com/api3dao/airseeker/pull/343/commits/42707a4b71fdaa34185da407ec5b340f695c0a9f

config/configuration.md
acenolaza: https://github.com/api3dao/airseeker/pull/343/commits/6c229e8809c7059eec4a92c184f9bd5ab48c91d3

src/update-feeds-loops/get-updatable-feeds.ts
acenolaza: https://github.com/api3dao/airseeker/pull/343/commits/8bc4864525ae2f8589db51bac277c79661c5b621

src/update-feeds-loops/get-updatable-feeds.ts
acenolaza: https://github.com/api3dao/airseeker/pull/343/commits/1bbf7970ff68b79cede4d5c919705992808a0fc6

config/configuration.md
acenolaza: https://github.com/api3dao/airseeker/pull/343/commits/cafec8dfc47943d323242247581cd33f0c17321e

src/update-feeds-loops/get-updatable-feeds.test.ts
acenolaza: https://github.com/api3dao/airseeker/pull/343/commits/f41fb2335b984b2f21cc4b8fa2b14b9fd191cb05

src/update-feeds-loops/get-updatable-feeds.ts
acenolaza: I think I do in the case where: 
1. data feed is a beacon (isBeaconSet is false)
2. isValidBeaconUpdate is true
3. checkUpdateCondition() returns false
4. individualBeaconUpdateDeviationThresholdCoefficient is set in the config

src/update-feeds-loops/get-updatable-feeds.ts
acenolaza: I think I need more details on what's too complex here. Are you saying that calling checkUpdateCondition() inside an Array.filter() then Array.map() is to complex or that we use a factor of the adjustedDeviationThresholdCoefficient instead of a hardcoded percentage is too complex? If it is the later, then I guess the idea is that individual beacon update requirement is more relaxed than the beacon set threshold. For example, if beaconSet should update at 1% deviation then individual beacon update should only update when 1% is not reached but some beacon has reached 5% deviation. If we make the config option a percentage then I guess we should validate at least that it is higher than adjustedDeviationThresholdCoefficient with zod

src/update-feeds-loops/pending-transaction-info.ts
acenolaza: Sorry, I'm confused. You mean `const isFeedUpdatable`? I've changed it because it is no longer a boolean

config/configuration.md
acenolaza: https://github.com/api3dao/airseeker/pull/343/commits/cafec8dfc47943d323242247581cd33f0c17321e

src/update-feeds-loops/pending-transaction-info.ts
acenolaza: They are also added to logger context on lines 77 and 92 so I'll keep them here

config/configuration.md
acenolaza: https://github.com/api3dao/airseeker/pull/343/commits/e1183d04a3651a1af27d8726e7812fa106a6b2bc

src/update-feeds-loops/submit-transactions.ts
acenolaza: https://github.com/api3dao/airseeker/pull/343/commits/831224d7c3131f9ef443db7139e98328774e4144

src/update-feeds-loops/get-updatable-feeds.test.ts
acenolaza: https://github.com/api3dao/airseeker/pull/343/commits/772561b804742077ae8c68243f7c42fa2f87ffbc

src/update-feeds-loops/get-updatable-feeds.ts
acenolaza: https://github.com/api3dao/airseeker/pull/343/commits/4089d86ea8283c4b34042d57ff37f8616b708c09

src/update-feeds-loops/get-updatable-feeds.ts
acenolaza: Really good catch 👏🏻 I guess we should only push to the accumulator if updatableBeacons.length > 0

acenolaza: https://github.com/api3dao/airseeker/pull/343/commits/b54cf23fc559cc23b4bd33c6cb5aac5170bc9536

acenolaza:
> Rather than extending `UpdatableDataFeed` with `shouldUpdateBeaconSet`, we may define `isAsyncBeaconUpdate`.

I don't have a strong opinion about which one is better. @Siegrift thoughts?

bdrhn9 (approved):
Looks good to me 🔥

src/update-feeds-loops/get-updatable-feeds.ts
```diff
@@ -19,122 +19,167 @@ export interface UpdatableBeacon {
...
+    const isBeaconSet = beaconValues.length > 1;
+
+    // We need to make sure that the update transaction actually changes the
+    // on-chain value. There are two cases:
+    // 1. Data feed is a beacon - We need to make sure the off-chain data
+    //    updates the feed. This requires timestamp to change. See:
+    //    https://github.com/api3dao/airnode-protocol-v1/blob/65a77cdc23dc5434e143357a506327b9f0ccb7ef/contracts/api3-server-v1/DataFeedServer.sol#L120
+    const isValidBeaconUpdate = !isBeaconSet && newDataFeedTimestamp !== dataFeedTimestamp;
+
+    // 2. Data feed is a beacon set - We need make sure that the beacon set will change. The contract requires the
```
Siegrift:
```suggestion
    // 2. Data feed is a beacon set - We need to make sure that the beacon set will change. The contract requires the
```

config/configuration.md
```diff
@@ -201,6 +201,24 @@ The global coefficient applied to all deviation checks. Used to differentiate al
 
 Defaults to `1`.
 
+#### `individualBeaconUpdateDeviationThresholdCoefficient`
+
+The global deviation factor applied to individual beacon updates deviation checks. When configured, this value will be
```
Siegrift:
```suggestion
The global deviation coefficient applied to individual beacon updates deviation checks. When configured, this value will be
```

src/update-feeds-loops/get-updatable-feeds.ts
```diff
@@ -19,122 +19,167 @@ export interface UpdatableBeacon {
...
           })),
-      }))
-  );
+        shouldUpdateBeaconSet: isBeaconSet,
+      });
+      continue;
+    }
+
+    if (isBeaconSet && individualBeaconUpdateDeviationThresholdCoefficient) {
+      // There is a special case when data feed is a beacon set that do not need
```
Siegrift:
```suggestion
      // There is a special case when data feed is a beacon set that does not need
```

src/update-feeds-loops/get-updatable-feeds.ts
```diff
@@ -19,122 +19,167 @@ export interface UpdatableBeacon {
...
+      // to be updated but some of its beacon constituents do. In this
+      // particular case, Airseeker can update only these beacons and skip the
+      // beacon set update. This is enabled by setting a value in
```
Siegrift: nit: ```suggestion
      // beacon set update.
```

I'd remove the last sentence.

src/update-feeds-loops/pending-transaction-info.ts
Siegrift: I see. Makes sense.

Siegrift (approved):
👍 LGTM

Siegrift:
> I don't have a strong opinion about which one is better. @Siegrift thoughts?

I don't particularly like the `shouldUpdateBeaconSet` but I would need to see how the code would look like with `isAsyncBeaconUpdate`. The problem I have with that field is that you also need to populate it for single beacon feeds, for which the field makes no sense (and you have to have an additional check in the if statement before changing the pending tx info).

acenolaza:
Thank you both for the valuable feedback 🙏🏻  I'm going to merge this now to be able to have this functionality ready. We can continue the discussion to refactor the `shouldUpdateBeaconSet` flag on a different issue

PR #348, opened 2024-07-11
Upgrade to pnpm v9

### LINKED ISSUE #342: update to pnpm@9
As per this [thread](https://api3workspace.slack.com/archives/C01N6001YCE/p1719829565121759)

dcroote:
Closes #342. The only difference between this upgrade and the same in other repos is I've set the pnpm version to match within the Docker file. I also got an error initially when running the docker container because it needs a `.env` file so I mention that in the README.


Siegrift (approved):
👍 LGTM

PR #361, opened 2024-08-01
Add package manager field

aquarat:
Relates to https://github.com/api3dao/tasks/issues/1338

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -21,15 +21,15 @@ jobs:
     steps:
       - name: Clone repo
         uses: actions/checkout@v4
-      - name: Install pnpm
-        uses: pnpm/action-setup@v3
+      - name: Enable Corepack
+        run: 'corepack enable'
       - name: Setup Node
         uses: actions/setup-node@v4
         with:
           node-version: '20.x'
           cache: 'pnpm'
       - name: Install Dependencies
-        run: pnpm install --frozen-lockfile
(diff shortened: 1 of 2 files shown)


bdrhn9 (approved):
Should we want to remove following lines: 

- https://github.com/api3dao/airseeker/blob/9e86c086af4a30da2f6afbcc89e269ed06ab8193/.github/workflows/main.yml#L27
- https://github.com/api3dao/airseeker/blob/9e86c086af4a30da2f6afbcc89e269ed06ab8193/.github/workflows/main.yml#L54

.github/workflows/main.yml
```diff
@@ -21,17 +21,15 @@ jobs:
     steps:
       - name: Clone repo
         uses: actions/checkout@v4
-      - name: Install pnpm
-        uses: pnpm/action-setup@v3
-        with:
-          version: 9
+      - name: Enable Corepack
+        run: 'corepack enable'
```
dcroote, line 25: @aquarat  can I ask why corepack instead of the pnpm action? With the pnpm action, the version is optional when `packageManager` is specified in package.json ([docs here](https://github.com/pnpm/action-setup/tree/v4/?tab=readme-ov-file#version)).

aquarat: Hey Derek
I'm only seeing this now. I didn't realise there was a pnpm action - if it's better, feel free to use that instead.

PR #377, opened 2024-09-03
Reduce info-level logging in data fetcher loop

### LINKED ISSUE #376: Reduce info-level logging in data fetcher loop
Currently, more than half of our logs are generated by the data fetcher loop ([link 1](https://github.com/api3dao/airseeker/blob/ca84adef4f01dd0b2630ec459efde199094ec8ee/src/data-fetcher-loop/data-fetcher-loop.ts#L108), [link 2](https://github.com/api3dao/airseeker/blob/ca84adef4f01dd0b2630ec459efde199094ec8ee/src/data-fetcher-loop/data-fetcher-loop.ts#L118)). While these logs provide valuable insights for debugging issues on the signed API side and verifying signed data before it is stored in Airseeker, they are contributing to excessive log volume. This clutter makes it challenging to manually track and analyze logs effectively.

To address this, I propose that we downgrade these logs from info to debug level. This change would still allow us to retain the ability to troubleshoot and verify data without overwhelming our logs.

Alternatively, we could implement a summary log at the end of each loop iteration. This summary could include statistics in the following format. By consolidating the information into a single log entry, we maintain almost the same level of visibility while significantly reducing log volume.

```typescript
{ 
averageFetchDuration: number,
averageSaveDuration: number,
fastestFetch: { url: string, duration: number },
slowestFetch: { url: string, duration: number },
fastestSave: { url: string, duration: number },
slowestSave: { url: string, duration: number }
}
```

bdrhn9:
Closes #376.

### DIFF
--- a/src/data-fetcher-loop/data-fetcher-loop.ts
+++ b/src/data-fetcher-loop/data-fetcher-loop.ts
@@ -1,5 +1,5 @@
 import { type Hex, executeRequest } from '@api3/commons';
-import { uniq } from 'lodash';
+import { minBy, meanBy, maxBy, uniq } from 'lodash';
 
 import { logger } from '../logger';
 import { getState } from '../state';
@@ -7,6 +7,12 @@ import { type SignedDataRecord, signedApiResponseSchema, type SignedDataRecordEn
 import { generateRandomId, sleep } from '../utils';
 
 import { purgeOldSignedData, saveSignedData } from './signed-data-state';
+
+interface SignedApiUrlStats {
+  url: string;
+  count: number;
+  duration: number;
+}
 
 export const startDataFetcherLoop = () => {
   const state = getState();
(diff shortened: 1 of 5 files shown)


src/data-fetcher-loop/data-fetcher-loop.ts
```diff
@@ -94,35 +100,51 @@ export const runDataFetcher = async () => {
...
-        });
+        const saveDuration = Date.now() - saveStart;
+        saveDurations.push({ url, count: signedDataCount ?? 0, duration: saveDuration });
       })
     );
 
+    // Log the statistics for the data fetcher loop.
+    logger.info('Finished data fetcher loop.', {
+      loopDuration: Date.now() - new Date(loopStartedAt).getTime(),
+      averageFetchDuration: fetchDurations.length === 0 ? undefined : meanBy(fetchDurations, 'duration'),
```
bdrhn9, line 140: I'm doing check for `.length === 0` in order to avoid divide by zero while calculating mean.

Siegrift (approved):
👍 LGTM

PR #420, opened 2024-12-09
Bump @api3/contracts & @api3/chains

bdrhn9:
- Bumped @api3/contracts and @api3/chains which requires attention because they includes major changes
- Removed @api3/airnode-protocol-v1 dependency and utilized @api3/contracts instead
- Regenerated pnpm-lock.yaml from scratch


hiletmis (approved):
LGTM. thanks for using up to date packages.

PR #450, opened 2025-03-03
Introduce heartbeat interval modifier for individual beacon updates

### LINKED ISSUE #449: Introduce heartbeat interval modifier for individual beacon updates
Related to https://github.com/api3dao/airseeker/issues/329.

The reasoning behind this implementation is to prevent redundant beacon updates just before an incoming heartbeat update for the beacon set. For example, in the case of a stablecoin, it is expected to have one update per day. Take USDC/USD as an example: when it is updated, its timestamp will be higher than half of its beacons due to the median calculation. In the next heartbeat update cycle, individual beacon updates will occur for the lagging beacons just before the actual beacon set update. Therefore, it would be better to introduce a heartbeat interval modifier for individual beacon updates, similar to the global heartbeat interval modifier.

bdrhn9:
Closes #449

### DIFF
--- a/config/configuration.md
+++ b/config/configuration.md
@@ -201,15 +201,24 @@ The global coefficient applied to all deviation checks. Used to differentiate al
 
 Defaults to `1`.
 
-#### `individualBeaconUpdateDeviationThresholdCoefficient`
+#### `individualBeaconUpdateSettings`
 
-The global deviation coefficient applied to individual beacon updates deviation checks. When configured, this value will
-be used on beacon sets that don't require an update but some of their beacon constituents do. For instance, if a beacon
-set has a 1% deviation threshold, Airseeker will update its beacons only when the deviation of each beacon exceeds (1 \*
-individualBeaconUpdateDeviationThresholdCoefficient)%. For example:
+The global settings applied to individual beacon updates. When configured, these settings will be used on beacon sets
+that don't require an update but some of their beacon constituents do. The object has the following properties:
+
+- `deviationThresholdCoefficient`: The deviation coefficient applied to individual beacon updates deviation checks. For
+  instance, if a beacon set has a 1% deviation threshold, Airseeker will update its beacons only when the deviation of
+  each beacon exceeds (1 \* individualBeaconUpdateSettings.deviationThresholdCoefficient)%.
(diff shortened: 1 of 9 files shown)


config/configuration.md
```diff
@@ -201,35 +201,44 @@ The global coefficient applied to all deviation checks. Used to differentiate al
...
+that don't require an update but some of their beacon constituents do. The object has the following properties:
 
-The global modifier applied to all heartbeat checks. The value is specified in seconds and can be positive or negative.
-The resulting interval is calculated by adding this modifier as an offset and clamped to a minimum of 0, ensuring it
-never goes negative. Used to differentiate alternate deployments. For example:
+- `deviationThresholdCoefficient`: The deviation coefficient applied to individual beacon updates deviation checks. For
+  instance, if a beacon set has a 1% deviation threshold, Airseeker will update its beacons only when the deviation of
+  each beacon exceeds (1 \* individualBeaconUpdateSettings.deviationThresholdCoefficient)%. Defaults to `1`.
+- `heartbeatIntervalModifier`: The modifier applied to the heartbeat interval for individual beacon updates. This can be
+  used to adjust the heartbeat interval for individual beacons. Defaults to `0`.
```
Siegrift, line 225: Shouldn't it default to something positive instead? Like 1-2 minutes?

src/update-feeds-loops/get-updatable-feeds.ts
```diff
@@ -76,7 +76,7 @@ export const getUpdatableFeeds = (
       const newDataFeedValue = calculateMedian(beaconValues.map(({ value }) => value));
       const newDataFeedTimestamp = calculateMedian(beaconValues.map(({ timestamp }) => timestamp));
 
-      const adjustedDeviationThresholdCoefficient = multiplyBigNumber(
+      const adjustedDeviationThreshold = multiplyBigNumber(
```
Siegrift: 👍

Siegrift (approved):
👍 LGTM. Thanks for splitting this into multiple commits 🙏

config/configuration.md
bdrhn9: I thought similarly at first, but then I considered that if a user specifies only the `deviationThreshold` and forgets to define the `heartbeatInterval`, Airseeker should probably use a neutral element for `heartbeatInterval`, which is 0.

PR #452, opened 2025-03-03
Toggle utilizing signed API URLs from AirseekerRegistry

### LINKED ISSUE #451: Add functionality to toggle utilizing signed API URLs from AirseekerRegistry
Implement configuration option to enable/disable the use of signed API URLs from AirseekerRegistry. This enhancement will help users with limited network bandwidth who prefer to rely exclusively on signed API URLs specified in their configuration.

bdrhn9:
Closes #451


config/configuration.md
```diff
@@ -252,6 +252,18 @@ The fetch interval in seconds between retrievals of signed API data.
 
 A list of signed API URLs to call along with URLs fetched from the chain.
 
+### `useSignedApiUrlsFromContract` _(optional)_
+
+A boolean flag that determines whether to use signed API URLs from the AirseekerRegistry contract. When set to `true`,
```
Siegrift:
```suggestion
A boolean flag that determines whether to use Signed API URLs from the AirseekerRegistry contract. When set to `true`,
```

src/data-fetcher-loop/data-fetcher-loop.test.ts
```diff
@@ -117,6 +118,76 @@ describe(dataFetcherLoopModule.runDataFetcher.name, () => {
...
+    expect(callSignedApiSpy).toHaveBeenCalledWith(
+      'http://127.0.0.1:8090/0xC04575A2773Da9Cd23853A69694e02111b2c4182',
+      10_000
+    );
+    expect(callSignedApiSpy).toHaveBeenCalledWith('http://127.0.0.1:8091/0xContractProvidedAirnode', 10_000);
+
+    expect(logger.debug).toHaveBeenCalledWith(
+      'Signed API URLs.',
+      expect.objectContaining({
+        useSignedApiUrlsFromContract: true,
```
Siegrift, line 185: nit: the `useSignedApiUrlsFromContract` is not really needed for the logging purposes given that we log `signedApiUrlsFromContract`. Maybe remove that property and instead assert the correctness of `useSignedApiUrlsFromContract`?

Siegrift (approved):
👍 LGTM

src/data-fetcher-loop/data-fetcher-loop.test.ts
bdrhn9: >  the `useSignedApiUrlsFromContract` is not really needed for the logging purposes given that we log `signedApiUrlsFromContract`.

Even `useSignedApiUrlsFromContract` is false, we are collecting `signedApiUrlsFromContract` [here](https://github.com/api3dao/airseeker/blob/f93eab27ece4fcc05782fb38e686ff9c966924fc/src/update-feeds-loops/update-feeds-loops.ts#L254). This flag only changes utilization of them. So logging `signedApiUrlsFromContract` doesn't tell whether they are used or not. This information can be removed but I'd prefer to keep because it is in debug level.

> instead assert the correctness of useSignedApiUrlsFromContract?

Doesn't `callSignedApiSpy` [expectations](https://github.com/api3dao/airseeker/blob/f93eab27ece4fcc05782fb38e686ff9c966924fc/src/data-fetcher-loop/data-fetcher-loop.test.ts#L180) meet this validation?

config/configuration.md
Siegrift: About the "signed API" vs "Signed API" we should always use latter, because the former one means "API that is signed" which doesn't make sense. The latter is clear in that it's a service name.

Siegrift (approved):

config/configuration.md
bdrhn9: You're right. I went over whole repository here [97345b7](https://github.com/api3dao/airseeker/pull/452/commits/97345b75b6ff0014fd410ea61e9dd3e212a74414).

PR #541, opened 2025-10-02
Update .dockerignore

Siegrift:
I've inspected the docker container of Airseeker and found multiple files that are not needed in the built container.


Siegrift:
To verify the `.dockerignore` changes, one can run:
1. pnpm run docker:build
2. docker run --rm -it --entrypoint /bin/sh api3/airseeker:latest
3. ls -lah

I have also run locally to see if it starts, but we should keep an eye on it when using the newly released image just in case.

bdrhn9 (approved):
Changes look good to me. However, it sounds like it would be better to update the Dockerfile to only copy the required parts to the container [here](https://github.com/api3dao/airseeker/blob/19bc8eb90daba699794a8cd6101a821188b0c3d0/docker/Dockerfile#L15), rather than filling `.dockerignore` which is hard to track.

Siegrift:
Yeah, although that has the same problem as you'll need to list what files/folders to include.

PR #570, opened 2025-12-15
Add `contents: write` permission to release workflow

bdrhn9:
### Problem
The release workflow failed during tag creation with the following error: https://github.com/api3dao/airseeker/actions/runs/20232703249/job/58123311904#step:10:29

This issue was unexpected as the workflow previously functioned without problems.

### Solution
After debugging in a personal repository, the issue was identified as missing `contents: write` permission. This permission has been added to the `tag-and-release` job to enable pushing tags to the repository.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -93,6 +93,7 @@ jobs:
         node-version: [20]
     permissions:
       id-token: write
+      contents: write
     steps:
       - name: Clone repo
         uses: actions/checkout@v5


metobom (approved):

Siegrift:
> This permission has been added to the tag-and-release job to enable pushing tags to the repository.

Mhh, but tag and release script has been unchanged for quite some time now and the releases worked fine without it 🤔 Btw. I think we have the same problem in other repos with the release workflow.

bdrhn9:
> > This permission has been added to the tag-and-release job to enable pushing tags to the repository.
> 
> Mhh, but tag and release script has been unchanged for quite some time now and the releases worked fine without it 🤔 Btw. I think we have the same problem in other repos with the release workflow.

I found the actual problem. The issue stems from [the change](https://github.com/api3dao/airseeker/pull/541/files) in `main.yml` that removed `token: ${{ secrets.GH_ACCESS_TOKEN }}` from `actions/checkout`. This token is necessary because it works similarly to `git clone https://${GH_USERNAME}:${GH_TOKEN}@github.com/api3dao/airseeker.git` where Git saves the token credentials for subsequent operations.

Later in the workflow, we define `GH_ACCESS_TOKEN` in `tag-and-release`, but that's only used for Octokit actions within the [createGithubRelease](https://github.com/api3dao/commons/blob/1dcd4c680045fab3e7d33759c30ecf1a01952259/src/release-scripts/tag-and-release.ts#L41) function. However, this function also executes native `git push` commands, which need the saved credentials. Without the token in `actions/checkout`, these `git push` operations fail which is why our CI failed yesterday. 

I've confirmed what I mentioned above by testing all alternatives in my test repository: https://github.com/bdrhn9/test-release.

PR #576, opened 2026-01-05
Update workflow script

Siegrift:
As part of NPM trusted publishing.


Siegrift:
I'm merging, because I've already performed this exact change on other repos and also plan to do a test release anyway.

PR #588, opened 2026-01-25
Make `terminateKeycardWallet` idempotent

### LINKED ISSUE #587: Make `terminateKeycardWallet` idempotent
Cleanup functions should be safe to call regardless of initialization state. This is especially important in shutdown handlers where graceful cleanup is needed without tracking whether each component was fully initialized. 

Additionally, regarding the code [here](https://github.com/api3dao/airseeker/blob/59510896fdbc0e1a8a7c1e33fa217fa420a62ef0/src/index.ts#L13), `terminateKeycardWallet` is currently called unconditionally during exit. This causes the program to throw an error if the Keycard wallet wasn't initialized. While this doesn't break the application, it creates unnecessary noise in the logs and should be handled more gracefully.

bdrhn9:
Closes #587.

### DIFF
--- a/src/keycard.ts
+++ b/src/keycard.ts
@@ -20,9 +20,8 @@ export const initializeKeycardWallet = async () => {
 };
 
 export const terminateKeycardWallet = () => {
-  if (!keycardWallet) {
-    throw new Error('Keycard wallet is not initialized.');
-  }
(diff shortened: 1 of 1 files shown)


Siegrift (approved):
👍 LGTM

